PR #3, opened 2018-07-12
Multithreaded scheduler

neeleshdodda44:
Comments in code are hopefully pretty self-explanatory, but to summarize + extra details:

- Using Ray to run runners asynchronously 
- `DMLRunnerActor` is just a wrapper for `DMLRunner` (didn't want to mess with the tests for `DMLRunner`, which would clearly fail if I annotated `DMLRunner` as the actor)
- Can run with as many runners as you want (default is 4), but ran into some weird errors when I was using too many with little speed boost. Feel free to change the default.
- The test for speedup is fairly naive (Georgy mentioned that the scheduler is throwaway, so I didn't want to test by hardcoding stuff). 
- New to pull requests so I don't know if I'm the one merging...? If not, I'm fairly sure you can just add the new test file, add the new actor class, and replace the scheduler code with mine (unless there were significant changes to `runner.py` or smt

PR #5, opened 2018-07-13
Multithreaded scheduler

georgymh:
From the reverted pull request #3:

> Comments in code are hopefully pretty self-explanatory, but to summarize + extra details:
> 
> - Using Ray to run runners asynchronously
> - DMLRunnerActor is just a wrapper for DMLRunner (didn't want to mess with the tests for DMLRunner, which would clearly fail if I annotated DMLRunner as the actor)
> - Can run with as many runners as you want (default is 4), but ran into some weird errors when I was using too many with little speed boost. Feel free to change the default.
> - The test for speedup is fairly naive (Georgy mentioned that the scheduler is throwaway, so I didn't want to test by hardcoding stuff).
> - New to pull requests so I don't know if I'm the one merging...? If not, I'm fairly sure you can just add the new test file, add the new actor class, and replace the scheduler code with mine (unless there were significant changes to runner.py or smt


core/runner.py
```diff
@@ -43,6 +43,7 @@ def run_job(self, job):
 
         If the runner is already executing a job, it silently does nothing.
         """
+
```
georgymh: Can you remove this line.

core/runner_actor.py
```diff
@@ -0,0 +1,102 @@
...
+    DML Runner Actor
+
+    NOTE: This is a wrapper around DMLRunner so that testing on DMLRunner can be 
+    independent of interaction with Ray
+
+    NOTE 2: Please read https://ray.readthedocs.io/en/latest/actors.html. This 
+    class serves as an actor, where methods of each instance are run in parallel 
+    across instances. This is so that runners can run jobs independently of each
+    other.
+
```
georgymh: No need for the docs in lines 35-43. We wouldn't want to maintain docs in the original class and the actor.

core/runner_actor.py
```diff
@@ -0,0 +1,102 @@
+# Add the parent directory to the PATH to allow imports.
```
georgymh: Why do we need lines 1-5? I think this is only useful for things inside `tests`. We're not running files directly and calling their main entrypoint anymore.

core/runner_actor.py
```diff
@@ -0,0 +1,102 @@
...
+import ray
+
+from core.runner import DMLRunner
+from custom.keras import model_from_serialized, get_optimizer
+from data.iterators import count_datapoints
+from data.iterators import create_train_dataset_iterator
+from data.iterators import create_test_dataset_iterator
+from core.utils.keras import train_keras_model, validate_keras_model
+from core.utils.keras import serialize_weights
+
```
georgymh: Extra line here. Our style guide is 2 lines between imports and code.

core/scheduler.py
```diff
@@ -3,24 +3,24 @@
...
 from core.utils.dmljob import DMLJob
-from core.runner import DMLRunner
+from core.runner_actor import DMLRunnerActor
 
+
+ray.init()
 logging.basicConfig(level=logging.DEBUG,
                     format='[Scheduler] %(asctime)s %(levelname)s %(message)s')
 
 class DMLScheduler(object):
```
georgymh: Write docstrings for each method. Some methods have comments right before their definition. These comments should be moved to the docstrings or inside the method definition.

tests/test_dmlscheduler.py
```diff
@@ -0,0 +1,93 @@
...
+import numpy as np
+from custom.keras               import get_optimizer
+from core.runner                import DMLRunner
+from core.runner_actor          import DMLRunnerActor
+from models.keras_perceptron    import KerasPerceptron
+from core.utils.dmljob          import DMLJob, serialize_job, deserialize_job
+from core.scheduler             import DMLScheduler
+from examples.labelers          import mnist_labeler
+from queue import Queue
+print("testing")
```
georgymh: Remove the printing statement here.

tests/test_dmlscheduler.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+import ray
+
+
+def make_dataset_path():
```
georgymh: You can remove this function since you're not using it.

tests/test_dmlscheduler.py
```diff
@@ -0,0 +1,93 @@
...
+    return 'datasets/mnist'
+
+
+def make_config():
```
georgymh: You can remove this function since you're not using it.

tests/test_dmlscheduler.py
```diff
@@ -0,0 +1,93 @@
...
+        scheduler._runners_run_next_jobs()
+    initial_weights = scheduler.processed.pop(0)
+    print(time.time()-start)
+    assert type(initial_weights) == list
+    assert type(initial_weights[0]) == np.ndarray
+    scheduler.reset()
+
+#Check that running 10 jobs with Ray takes at most half the 
+# time to run them serially (i.e. there is a speedup)
+def test_dmlscheduler_speedup_naive():
```
georgymh: I don't think this is a good test. It's a good proof to justify why this scheduler is better, but I don't think it's something that should be run every time. It also depends on your hardware resources. I would remove this.

tests/test_dmlscheduler.py
```diff
@@ -0,0 +1,93 @@
...
+def test_dmlscheduler_sanity():
+    model_json = make_model_json()
+    initialize_job = make_initialize_job(model_json)
+    scheduler.add_job(initialize_job)
+    scheduler._runners_run_next_jobs()
+    start = time.time()
+    while not scheduler.processed:
+        scheduler._runners_run_next_jobs()
+    initial_weights = scheduler.processed.pop(0)
+    print(time.time()-start)
```
georgymh: Remove printing statement here. We're more interested in this not timing out rather than how long it takes in this stage (code).

tests/test_dmlscheduler.py
```diff
@@ -0,0 +1,93 @@
...
+    initialize_job = DMLJob(
+        "initialize",
+        model_json,
+        "keras"
+    )
+    return initialize_job
+
+scheduler = DMLScheduler()
+
+#Check that the scheduling/running functionality is maintained.
```
georgymh: Put this comment inside as a docstring.

tests/test_dmlscheduler.py
```diff
@@ -0,0 +1,93 @@
...
+    start = time.time()
+    while len(scheduler.processed) < 10:
+        scheduler._runners_run_next_jobs()
+    fast = time.time() - start
+    print(slow)
+    print(fast)
+    assert fast*1.5 < slow
+    scheduler.reset()
+
+
```
georgymh: Can we add a test that checks if 5 or 10 things can be scheduled at different times (for example with five things: scheduling 2 at the beginning, 2 after the 1st executes, one after the 4th executes) and verify that all jobs were executed? 

Also a test that checks if jobs are picked up and executed regardless of the scheduler being interacted with (basically that line 81 in the `scheduler.py` works)? (This would need to run fast, so the `run_frequency_in_mins` should be set to something like 5 seconds to keep the whole test under 10 seconds.)

georgymh: Overall okay, but there are some changes to be made.

georgymh:
@neeleshdodda44: Please see the comments above.

core/runner.py
neeleshdodda44: Done

core/runner_actor.py
neeleshdodda44: Removed lines 35-43. The method comments are copied exactly from `DMLRunner`, are you sure you still want me to delete them?

core/runner_actor.py
neeleshdodda44: Copied them exactly from `DMLRunner` and accidentally left them in. Removed.

core/runner_actor.py
neeleshdodda44: Done.

core/scheduler.py
neeleshdodda44: Done.

tests/test_dmlscheduler.py
neeleshdodda44: Done.

tests/test_dmlscheduler.py
neeleshdodda44: Removed.

tests/test_dmlscheduler.py
neeleshdodda44: Removed.

tests/test_dmlscheduler.py
neeleshdodda44: I completely agree with your logic, and I actually thought about this problem for a while. Even though the test is naive (as I mentioned before), I still think it's necessary to have it. My reasoning is that if someone in the future messes with the class and accidentally breaks the threading and not the core functionality (i.e. every runner runs the same job), then we won't know since the tests would still pass.  I can always just lower the bar of the test if you're afraid of varying hardware resources causing occasional fails. 

If you have faith that the situation I described won't occur, though, I'll go ahead and delete the test.

tests/test_dmlscheduler.py
neeleshdodda44: Done

tests/test_dmlscheduler.py
neeleshdodda44: Done

tests/test_dmlscheduler.py
neeleshdodda44: The first test sounds good, I'll implement that.

As for the second test: since the CRON job method is something you wrote, I had a quick question. When the CRON job is set up, does it asynchronously call the method I specified every period (so the proper way to test would be to wait a few periods and then check if all jobs are done running)?

core/runner_actor.py
```diff
@@ -24,14 +25,6 @@ class DMLRunnerActor(DMLRunner):
     """
     DML Runner Actor
 
-    NOTE: This is a wrapper around DMLRunner so that testing on DMLRunner can be
```
georgymh: I meant to delete the lines with "This class manages the training...This class only executes the jobs." which are now lines 28-35. The lines with the NOTEs were actually useful. Can you bring those lines back and delete 28-35?

Also, yes, even though you copied the method docstrings directly from DMLRunner, I believe we should remove them and just put:

```
"""See method docstring in the original class for more information."""
```

That way we do not maintain documentations in parallel across files.

core/scheduler.py
```diff
@@ -57,17 +57,20 @@ def __init__(self, num_runners=4):
...
         self.queue.append(dml_job) 
 
     def create_runner(self):
+        """ Creates new runner. """
         return DMLRunnerActor.remote(self.dataset_path, self.runner_config)
 
-    #Check each job to see if it has a job running. If not, have the runner
-    # run the next job on the queue asynchronously of the others and collect 
-    # the result of the job that was running before (if applicable)
+    #
```
georgymh: Remove this comment.

core/scheduler.py
```diff
@@ -77,13 +80,15 @@ def _runners_run_next_jobs(self):
...
 
     def _start_cron(self, run_frequency_mins):
+        """ CRON job to run next jobs on runners, if applicable. """
         logging.info("Starting cron...")
         schedule.every(run_frequency_mins).minutes.do(self._runners_run_next_jobs)
         schedule.run_pending()
         logging.info("Cron started!")
 
-    #scheduler is a singleton, so need a way "start over"
+    #
```
georgymh: Remove this comment.

core/scheduler.py
```diff
@@ -77,13 +80,15 @@ def _runners_run_next_jobs(self):
...
     def reset(self):
+        """scheduler is a singleton, so need to get a new instance inplace """
         self.queue = deque()
         self.processed = []
```
georgymh: Remove the commented code below this line.

core/scheduler.py
```diff
@@ -57,17 +57,20 @@ def __init__(self, num_runners=4):
...
         return DMLRunnerActor.remote(self.dataset_path, self.runner_config)
 
-    #Check each job to see if it has a job running. If not, have the runner
-    # run the next job on the queue asynchronously of the others and collect 
-    # the result of the job that was running before (if applicable)
+    #
     def _runners_run_next_jobs(self):
+        """ Check each job to see if it has a job running. If not, have the runner run the 
+        next job on the queue asynchronously of the others and collect the result of the job
+        that was running before (if applicable) """
```
georgymh: This may be small, but can you follow [PEP 257](https://www.python.org/dev/peps/pep-0257/) and use new lines instead of trailing spaces at the beginning and end of all multi-line docstrings?

tests/test_dmlscheduler.py
```diff
@@ -1,7 +1,6 @@
 import pytest
```
georgymh: There are still printing statements in this file to remove.

core/scheduler.py
```diff
@@ -32,51 +32,66 @@ def getInstance():
             DMLScheduler()
         return DMLScheduler.__instance
 
-    def __init__(self):
+    #Default number of runners that are running at a time is 4
```
georgymh: These kind of comments are better inside docstrings rather than outside methods. But I actually think we should remove it since the code is pretty much self-documenting at the function definition.

tests/test_dmlscheduler.py
georgymh: Ok. Once the first test is implemented and the minor things are changed we will be ready to merge.

For the second one, can you check [the documentation for the `schedule` module](http://schedule.readthedocs.io/en/stable/)? I found the module by googling but I wasn't able to test it until the multithreading capabilities were in place. I think you're correct on your approach to test it. We're not locked into the module, so if it's too hard to set up you can change it. We just want things to run as long as there are things to be run (as long as things are in the queue).

neeleshdodda44: Completed both tests. 

Messaged you on Slack regarding the issues of `schedule` and `ray` altogether in the process of making the second test, and why I ended up using `multiprocessing`

core/runner_actor.py
neeleshdodda44: Decided to move away from `ray`, so no need for the actor class anymore. 

Tl;dr resolved.

core/scheduler.py
neeleshdodda44: Removed.

core/scheduler.py
neeleshdodda44: Removed.

core/scheduler.py
neeleshdodda44: Removed.

core/scheduler.py
neeleshdodda44: Updated.

tests/test_dmlscheduler.py
neeleshdodda44: Removed (sorry I missed a few from debugging).

core/scheduler.py
neeleshdodda44: Removed.

kiddyboots216:
@georgymh fyi i'm not approving PRs from now on so if this is good you should merge it

PR #6, opened 2018-07-18
cleaned iterator, made integer labeler

neeleshdodda44:
Two major changes:

1. Labeler is now an integer. So the `_create_dataset_iterator` method in `iterators.py` only accepts integers in the `labeler` argument. I changed the tests for runner so that the `make_train_job` and `make_validate_job` methods made jobs whose labelers were 0.

2. Cleaned up the `_create_dataset_iterator' so that it uses dataframes in processing the data. Hopefully it's a bit easier to read now?

Suggestions: 

1. Would it be easier if the labeler was actually a string (that denotes the column name of the data)? Not trying to nitpick, but wouldn't you have to know the column name before you label the data? It seems like it would make everything a lot more readable and debugging would be smoother. The code change on my end would be minimal.

Hopefully this is pretty straightforward and this can get merged quickly.


data/iterators.py
```diff
@@ -52,33 +54,26 @@ def _create_dataset_iterator(dataset_path, max_count, iter_type, batch_size, lab
...
-                    batch = []
-                line = labeler(line) if labeler else line
-                batch.append(line)
-                count += 1
+        file = pd.read_csv(full_path).head(total_points)
+        assert labeler >= 0 and labeler < len(file.columns)
+        column = list(file.columns)[labeler]
+        X, y = (file.drop(column, axis=1), pd.DataFrame(file[column]))
+        X_split, y_split = np.split(X, num_batches), np.split(y, num_batches)
+        y_split = [keras.utils.to_categorical(y, num_classes=10) for y in y_split]
```
georgymh: This shouldn't be here -- we can't hardcode the label conversion for MNIST because this function needs to be general for all datasets. The goal is that EDA will happen first, then transformations, and by the time this DML code is executed `y_split` will be exactly what the model needs as an input to do supervised training. 

The easiest fix would be to remove this line and change the optimizer of `model/keras_perceptron.py` to be `sparse_categorical_crossentropy`, so it can accept integers as labels instead of a one-hot vector. [This is the line.](https://github.com/georgymh/decentralized-ml/blob/master/models/keras_perceptron.py#L39)

data/iterators.py
```diff
@@ -52,33 +54,26 @@ def _create_dataset_iterator(dataset_path, max_count, iter_type, batch_size, lab
...
-                        X_list, y_list = list(), list()
-                        for x, _ in batch: X_list.append(x)
-                        for _, y in batch: y_list.append(y)
-                        yield (np.array(X_list), np.array(y_list))
-                    batch = []
-                line = labeler(line) if labeler else line
-                batch.append(line)
-                count += 1
+        file = pd.read_csv(full_path).head(total_points)
+        assert labeler >= 0 and labeler < len(file.columns)
```
georgymh: Can you add an assert message here?

tests/test_dmlrunner.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 import numpy as np
 
+
```
georgymh: No extra line here.

data/iterators.py
neeleshdodda44: Sorry, should have asked about this. I only included that line because I wasn't passing the runner tests without it, and I didn't want to mess with the tests.

Fixed.

data/iterators.py
neeleshdodda44: Done.

tests/test_dmlrunner.py
neeleshdodda44: Removed.

neeleshdodda44:
@georgymh Incorporated the changes you asked for. Couple things:

> If you can modify the 2 MNIST CSVs accordingly

Since you didn't explicitly tell me more information, I assumed we were talking about _the_ MNIST dataset, in which case I followed the schema here: https://www.kaggle.com/c/digit-recognizer/data
Let me know if this is not the case.

Also, where in the wiki would you like me to make the note about the CSVs?

PR #9, opened 2018-09-04
Dataset manager edits

neeleshdodda44:
Really simple pull request. Updated Dataset Manager to copy over directories into transformed folder. File directory would look something like this:

```
main/
        dataset1/
            dataset1.csv
            md_dataset1.csv
        dataset2/
            dataset2.csv
            md_dataset2.csv
        transformed/
            dataset1/
                06/10/16sgf.csv
            dataset2/
                06/10/16mlf.csv
```

Tl;dr one can now view the transformed datasets of the corresponding folders in the original data directory.

### DIFF
--- a/core/dataset_manager.py
+++ b/core/dataset_manager.py
@@ -12,7 +12,6 @@ import tests.context
 import numpy as np
 import pandas as pd
 
-#import core.utils.context
 from core.configuration import ConfigurationManager
 
 
--- a/core/dataset_manager.py
+++ b/core/dataset_manager.py
@@ -132,8 +132,10 @@ class DatasetManager():
             timestamp = str(datetime.datetime.now())
             r_string = random_string(5)
             new_name = timestamp + r_string
-            transformed_data.to_csv(os.path.join(new_folder, new_name + '.csv'),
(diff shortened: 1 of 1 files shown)


core/dataset_manager.py
```diff
@@ -118,11 +125,14 @@ def random_string(length):
...
         for name,data in raw_data.items():
+            new_folder = os.path.join(self.tfp, name)
+            if not os.path.exists(new_folder):
+                os.makedirs(new_folder)
             transformed_data = transform_function(data)
             timestamp = str(datetime.datetime.now())
             r_string = random_string(5)
             new_name = timestamp + r_string
-            transformed_data.to_csv(os.path.join(self.tfp, new_name + '.csv'),
+            transformed_data.to_csv(os.path.join(new_folder, new_name + '.csv'),
```
georgymh: Can you make this:

```
transformed_data.to_csv(
  os.path.join(new_folder, new_name + '.csv'),
  index=False
)
```

neeleshdodda44: Nice catch, will do.

PR #10, opened 2018-09-07
(Naive) DatasetClassifier + DBClient + DatasetManager edits

neeleshdodda44:
3 major changes: 

**(Naive) DatasetClassifier**: 
- Sets up framework for what classification will look like when we actually have a trained classifier
- For now, can be used to manually add labels to datasets (or just use DB client, TBD)
- Didn't write tests since it's not the long term option and the main functionality (posting to DB) is tested under DBClient
- Eventually will be called by DatasetManager after transform is completed

**DBClient**
- Label datasets and retrieve datasets with corresponding category
- Needed here so that DL2 Notebook can retrieve labels
- Unsure whether DL2 Notebook will use this DBClient (through Virtual Worker instance of UNIX Service) or have its own instance, TBD
- Will most likely replace RDS DB with DynamoDB for performance reasons, but I'll figure that out after MVP

**DatasetManager**
- Updated file directories so that dataset folders in transformed folder don't have `.csv` appended to them.

Roast away :)

### DIFF
--- a/core/database_config.json
+++ b/core/database_config.json
@@ -1,0 +1,8 @@
+{
+    "user": "datashark",
+    "pw": "datashark",
+    "db": "datasharkdb",
+    "host": "datasharkdatabase.cwnzqu4zi2kl.us-west-1.rds.amazonaws.com",
+    "port": "5432",
+    "table_name": "category_labels"
+}
--- a/core/dataset_classifier.py
+++ b/core/dataset_classifier.py
@@ -1,0 +1,11 @@
+import pandas as pd
+
+
+class DatasetClassifier(object):
+	def __init__(self, label):
+		self.label = label
+
+	def classify_dataset(self):
+		if not self.label:
+			raise Exception('Classifier is not working at this time.')
+		return self.label
--- a/core/dataset_classifier.py
+++ b/core/dataset_classifier.py
@@ -1,11 +1,19 @@
 import pandas as pd
 
+from core.db_client import DBClient
+
 
 class DatasetClassifier(object):
-	def __init__(self, label):
+	def __init__(self, label=None):
 		self.label = label
+		self.db_client = DBClient()
 
-	def classify_dataset(self):
+	def _classify_dataset(self):
 		if not self.label:
 			raise Exception('Classifier is not working at this time.')
 		return self.label
+
+	def label_dataset(self, name):
+		label = self._classify_dataset()
+		self.db_client.add_labels([name], [label])
+
--- a/core/db_client.py
+++ b/core/db_client.py
@@ -1,0 +1,47 @@
+from flask import Flask
+from flask_sqlalchemy import SQLAlchemy
+import json
+import pandas as pd
+
+
+class DBClient(object):
+
+	def __init__(self, config_filepath = 'database_config.json'):
+		"""
+		Set up DBClient with corresponding database credentials
+		"""
+		app = Flask(__name__)
+		with open(config_filepath) as f:
+			POSTGRES = json.load(f)
+		app.config['SQLALCHEMY_DATABASE_URI'] = 'postgresql://%(user)s:%(pw)s@%(host)s:%(port)s/%(db)s' % POSTGRES
+		app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
+		self.db = SQLAlchemy(app)
+		self.table_name = POSTGRES['table_name']
+		self._reset()
+
+	def get_labels(self):
+		"""
+		Get category_labels table
+		"""
+		return pd.read_sql_query("select * from {table_name}".format(table_name=self.table_name), self.db.engine)
+
+	def add_labels(self, data_providers, categories):
+		"""
+		Append category labels to the category_labels table, where categories[i] corresponds to the labeled category
+		for data_providers[i]
+		"""
+		rows = {'data_provider': data_providers, 'category': categories} 
+		label = pd.DataFrame(data=rows)
+		label.to_sql(name=self.table_name, con=self.db.engine, if_exists='append', index=False)
+
+	def get_data_providers_with_category(self, category):
+		"""
+		Get a list of data providers with the given category.
+		"""
+		return pd.read_sql_query("select * from {table_name} where category = '{category}'"
+			.format(category=category, table_name=self.table_name), self.db.engine)
+
+	def _reset(self):
+		label = pd.DataFrame(columns=['data_provider', 'category'])
+		label.to_sql(name=self.table_name, con=self.db.engine, if_exists='replace', index=False)
+		
--- a/core/db_client.py
+++ b/core/db_client.py
@@ -22,21 +22,28 @@ class DBClient(object):
 		"""
 		Get category_labels table
 		"""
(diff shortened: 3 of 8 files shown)


kiddyboots216:
@neeleshdodda44 Travis CI build failed btw

core/database_config.json
```diff
@@ -0,0 +1,8 @@
+{
```
georgymh: I don't think showing/pushing passwords in plain-text is a good idea 😅 

One idea is to add this config to the Configuration Manager's setup and make the user enter it manually. I already added a `.gitignore` entry for the `core/configuration.ini` so we won't push passwords like this by mistake in the future -- we can have our own `.ini` that we use locally but never commit to version control.

As for the default value, we can leave it as empty for now. We may have to look into using Environment Variables in the future and/or making setup questions that are **required** (i.e., user won't be able to press enter and dismiss the question and the Configuration Manager won't continue until that question is answered).

Also, make sure to change the password of the DB if this is the actual password 🙂

core/dataset_classifier.py
```diff
@@ -0,0 +1,22 @@
+import pandas as pd
+
+from core.db_client import DBClient
+
+
+class DatasetClassifier(object):
+	
+	def __init__(self):
+		"""
+		Initalizes instance of Dataset Classifier. Only
```
georgymh: Is this doc unfinished?

core/dataset_classifier.py
```diff
@@ -0,0 +1,22 @@
...
+from core.db_client import DBClient
+
+
+class DatasetClassifier(object):
+	
+	def __init__(self):
+		"""
+		Initalizes instance of Dataset Classifier. Only 
+		"""
+		self.db_client = DBClient()
```
georgymh: I think it's a better idea to use dependency injection here as there's a parameter that can be set on the DBCLient class. It will be useful if we ever need to write tests that test the DatasetClassifier with a test DBClient.

core/db_client.py
```diff
@@ -0,0 +1,53 @@
...
+			labels.loc[data_provider] = category
+		labels['data_provider'] = labels.index
+		labels.index = list(range(len(labels.index)))
+		labels.to_sql(name=self.table_name, con=self.db.engine, if_exists='replace', index=False)
+
+	def get_data_providers_with_category(self, category):
+		"""
+		Get a list of data providers with the given category.
+		"""
+		self.get_labels()
```
georgymh: This returns something but it's not being captured.

core/db_client.py
```diff
@@ -0,0 +1,53 @@
...
+		"""
+		labels = pd.read_sql_query("select * from {table_name}".format(table_name=self.table_name), self.db.engine)
+		return labels
+
+	def add_labels(self, data_providers, categories):
+		"""
+		Append new category labels to the category_labels table and replace old labels, where categories[i] corresponds 
+		to the labeled category for data_providers[i]
+		"""
+		labels = pd.read_sql_query("select * from {table_name}".format(table_name=self.table_name), self.db.engine)
```
georgymh: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Use `self. get_labels()`.

georgymh: Also, what happens when getting the labels fails?

core/db_client.py
```diff
@@ -0,0 +1,53 @@
+from flask import Flask
+from flask_sqlalchemy import SQLAlchemy
+import json
+import pandas as pd
+
+
+class DBClient(object):
```
georgymh: Please write a docstring for the class.

Also, please put a note that authentication is not implemented yet so at the moment any data provider can change the labels for any data provider. We need to fix this later.

core/dataset_classifier.py
```diff
@@ -0,0 +1,22 @@
+import pandas as pd
+
+from core.db_client import DBClient
+
+
+class DatasetClassifier(object):
```
georgymh: Please write a docstring for the class.

core/db_client.py
```diff
@@ -0,0 +1,53 @@
...
+
+	def get_data_providers_with_category(self, category):
+		"""
+		Get a list of data providers with the given category.
+		"""
+		self.get_labels()
+		return pd.read_sql_query("select * from {table_name} where category = '{category}'"
+			.format(category=category, table_name=self.table_name), self.db.engine)
+
+	def reset(self):
```
georgymh: Please write a doctring for this function.

Also, why would we ever need to drop the whole table? This seems like a dangerous function to have.

tests/test_dbclient.py
```diff
@@ -0,0 +1,46 @@
...
+from core.db_client import DBClient
+
+@pytest.fixture
+def db_client():
+	"""
+	Maintain instance of DB Client
+	"""
+	return DBClient(config_filepath='tests/artifacts/db_client/database_config.json')
+
+def check_empty_category_labels(db_client):
```
georgymh: I'm not sure if I'm missing something (not sure if there's 1 database for each data provider or 1 for all), but when would the table be empty other than the moment when we have 0 active users?

core/database_config.json
neeleshdodda44: Oops I totally forgot this repo was public -_-

Will address!

neeleshdodda44: So a followup problem would be that the tests I wrote for DB Client use the DB. As of now, I have separate credentials in the `tests/artifacts/db_client` directory. So I propose two potential solutions:

1. Remove tests for DB Client. If we can't publish the password to version control, then there isn't any way for the tests to access the DB, right?

2. Modify tests to mock out DB. This would mean just having the table pulled locally as a csv so that the "get data provider by category" functionality can be tested.

Ideally, 2 would be great. But I also think it's overkill since most of the DB Client's functionality is running SQL commands on the DB. 

What do you think?

core/dataset_classifier.py
neeleshdodda44: No, accidentally typed `Only` there. Deleted.

core/dataset_classifier.py
neeleshdodda44: I agree. Updated.

core/db_client.py
neeleshdodda44: `self.get_labels()` is outdated, forgot to delete.

Deleted.

core/db_client.py
neeleshdodda44: Updated.

Getting the labels should only fail when the DB is down (and that should be a rare occurrence). Should I have error handling for that?

core/db_client.py
neeleshdodda44: Added docstring.

Could you clarify what you mean by:

> any data provider can change the labels for any data provider

We manage access to DB, and each data provider can only see the data they're giving us...?

core/dataset_classifier.py
neeleshdodda44: Added.

core/db_client.py
neeleshdodda44: Added docstring.

This function is strictly for the test table. Realizing that, I decided to move this function to the tests rather than the class.

neeleshdodda44:
@georgymh I replied to your comments. Just need to decide whether/how to test the DB Client. Also, could you explain what you mean by double tab indentation vs single tab?

PR #13, opened 2018-09-14
Remove scheduler speedup test

neeleshdodda44:
It's too hard to test speedup, and this test fails intermittently (out of our control).


georgymh:
LGTM

PR #14, opened 2018-09-16
Fixing train-test partitioning

### LINKED ISSUE #11: Fix train-test partitioning on data iterators
The data iterator in `data/iterators.py` currently partitions the dataset in a very naive way: it reverses the list of `.csv`'s inside the dataset folder and gets the top `floor(max_count / batch_size) * batch_size` datapoints of each `.csv` and yields them in batches.

This is incorrect because we instead need to yield a total of `max_count` datapoints across all `.csv`'s and because the sets of training data and testing data currently have an overlap (they should form a partition over the entire dataset, with the exception of the occasional data points that can't form a batch of size `batch_size`).

neeleshdodda44:
*This pull request serves to close #11*

Key changes:
- Changed pandas implementation back to original implementation, because the pandas implementation won't scale if dataset files have millions of rows.
- Fixed the underlying issue of overlapping data points. If the training dataset was obtained by iterating each dataset in order from top to bottom, then the test dataset should be obtained by iterating each dataset in reverse order from bottom to top (bug was test dataset iterating from top to bottom too).
- We are now using an integer labeler, but the standard of having a header should remain.


data/iterators.py
```diff
@@ -42,6 +43,7 @@ def create_test_dataset_iterator(dataset_path, count, split=0.2, batch_size=1, \
 
     Assumes `dataset_path` is a path to a folder with multiple CSV files.
     """
+
```
georgymh: Please remove this extra line.

core/db_client.py
```diff
@@ -49,6 +49,7 @@ def _get_labels(self):
 			try:
 				return pd.read_sql_query(query, self.db.engine)
 			except Exception as e:
+				print(e)
```
georgymh: Please log this exception instead.

tests/test_train_test_split.py
```diff
@@ -0,0 +1,63 @@
+from data.iterators import count_datapoints, create_train_dataset_iterator, create_test_dataset_iterator
```
georgymh: Can we test this on datasets where the number of datapoints % batch_size != 0, where batch_size % 2 != 0, batch_size > number of datapoints, *infinite = True*, labeler out of bounds, invalid batch_size, invalid dataset_path? If you think of other edge cases please add them too.

data/iterators.py
```diff
@@ -97,26 +104,31 @@ def _create_dataset_iterator(dataset_path, max_count, iter_type, batch_size, lab
         directories = reversed(os.listdir(dataset_path))
 
     count = 0
+    num_batches = 0
```
georgymh: This is not being used.

core/db_client.py
neeleshdodda44: Forgot to delete actually, but sure I'll actually just log it.

data/iterators.py
neeleshdodda44: Done.

data/iterators.py
neeleshdodda44: Was using for debugging, deleted.

tests/test_train_test_split.py
neeleshdodda44: I get all of these but why _infinite = True_? As far as I understand, it just wraps the generator in a while loop (the test would run the same, except I need to break out of the iteration when collecting data points).

neeleshdodda44: Also, why test invalid dataset_path? Isn't it the responsibility of the DatasetManager to handle that case?

neeleshdodda44: Tested the other things you mentioned (via the main test, or created new ones).

georgymh: For *infinite = True*, you can verify that we actually go through the data a finite amount of times (3N for example). I don't think we could check it's actually infinite.

True about the *dataset_path*, but we should make both components robust against invalid paths. The test should be in both.

data/iterators.py
```diff
@@ -98,15 +97,17 @@ def _create_dataset_iterator(dataset_path, max_count, iter_type, batch_size, lab
...
     if iter_type == 'train':
         directories = os.listdir(dataset_path)
     elif iter_type == 'test':
         directories = reversed(os.listdir(dataset_path))
 
     count = 0
-    num_batches = 0
     batch = []
     for filename in directories:
+        print(count)
```
georgymh: Please remove this line.

georgymh:
Test.

tests/test_train_test_split.py
```diff
@@ -0,0 +1,176 @@
...
+	Same as above, except test with batch_size > count (datapoints)
+	"""
+
+	count = count_datapoints(dataset_path)
+
+	#Set up iterator for training set
+	train_iterator = create_train_dataset_iterator(
+			dataset_path,
+			count,
+			batch_size=2000,
```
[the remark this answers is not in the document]
neeleshdodda44: Done for both. I just did `count * random.randint(1,5)` (to be simple, and I think your point was to just have a batch size that's always greater than the number of data points).

tests/test_train_test_split.py
neeleshdodda44: Added both!

data/iterators.py
neeleshdodda44: Done.

tests/test_train_test_split.py
```diff
@@ -0,0 +1,176 @@
...
+			labeler=-1,
+			infinite=False 
+		)
+
+	#Set up iterator for test set.
+	test_iterator = create_test_dataset_iterator(
+			dataset_path,
+			count,
+			batch_size=7,
+			labeler=-1,
```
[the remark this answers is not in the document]
neeleshdodda44: Sure, I made `labeler=2*num_columns`.

PR #16, opened 2018-09-22
Fix scheduler bug

neeleshdodda44:
_This pull request serves to close #12_

Should be a relatively straightforward pull request.

**Runner:**
- Error or not, `run_job` returns a dictionary called `results` with key `successful`
- If no error has occurred (`successful==True`), then `results` will have a key called `return_obj` with the value being results of the job run (`return_obj` is what was originally returned instead of `results`)
- Added error handling in `run_job`. If an error has occurred (`successful==False`), then `results` will have a key called `error_message` that will hold the actual error that took place when running the job.

**Scheduler:**
- Added error handling + retry logic in `runners_run_next_jobs`. If the job finished unsuccessfully, add it to the end of the queue. If the job finished unsuccessfully more than `max_tries` (currently 3) times, ignore the job and don't add it back to the queue (might need an alternative for this)
- `max_tries` is pulled from the `configuration.ini` 

**Other:**
- Edited `tests/artifacts/configuration.ini` and `questions.csv` to include `max_tries`.

### DIFF
--- a/core/runner.py
+++ b/core/runner.py
@@ -106,7 +106,6 @@ class DMLRunner(object):
             print(e)
         self.current_job = None
         logging.info("Finished running job!")
-        print(results)
         return results # Returning is only for debugging purposes.
 
     def _train(self, serialized_model, model_type, initial_weights, hyperparams,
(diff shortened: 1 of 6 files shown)


core/runner.py
```diff
@@ -50,46 +50,64 @@ def run_job(self, job):
...
+        except Exception as e:
+            results = {
+                "successful": False,
+                "error_message": e
+            }
+            print(e)
         self.current_job = None
         logging.info("Finished running job!")
-        return return_obj # Returning is only for debugging purposes.
+        print(results)
```
georgymh: Please remove this line.

PR #18, opened 2018-09-24
P2P and IPFS APIs

nzoghb:
Briefly skim and review when I get the tests up. Also will need a refactor for `runner.py` and `scheduler.py`

There's an alternative Kademlia module that has native callback support. Will look into after tests.


georgymh:
Closing for now since the Blockchain Client will be a superset of these utils.

PR #20, opened 2018-09-29
Adding Default Transform Folder

neeleshdodda44:
Major changes:
- On instantiation, the Dataset Manager creates a `transformed` folder that simply has the raw data.
- No more error handling when trying to get transformed data before transform (this is simply the raw data).
- Tests changed to reflect this new functionality.
- Customized exceptions are no longer necessary.


neeleshdodda44:
Addressing your comments:

- Yes, that was intended. I presumed that if there was going to be a identity transform at the start (even if it session vs initialization), then it might just be better to maintain that identity transform as the default _transformed_ data. I thought that it would be cleaner to just add transforms if necessary and directly pass the data (I was assuming that transforms/passing happens in one step) without having to do error handling to check for the transformed folder. Let me know what you think.
 
- Makes sense, but I do also want to point out some things. I'm not sure if start of session implies that the Config Manager is already set up, but why not just initialize the DM at the start of a session? Transforming at the start of a session will require a transform function and data filepath. So calling the transform function without initialization (I presume you mean something like a static method) actually seems less clean to me. Again, let me know what you think. 

- Sounds good, I can take up it as a task for this week.

PR #21, opened 2018-10-05
Secret Config

neeleshdodda44:
For sake of ease, define _secret sections_ as config sections that the user should not touch. Meaning, the user should not touch any entry in the section. Based on the modules we have, I assumed that sections were either entirely secret or entirely customizable (i.e. Scheduler is customizable, DBClient is secret). If this doesn't hold true, I can easily switch to commenting on each entry that the user shouldn't touch (although it would look a lot more messy).

Major changes:
- DBClient now uses Configuration Manager and pulls from `configuration.ini` to get its necessary configurations. I uncommented and ran its tests as a sanity check that it still works.
- For configurations that belong in a secret section, the corresponding entry for `question` is `SECRET`
- The Configuration Manager then filters entries in `questions.csv` based on whether the value of question is `SECRET` or not. If the value is `SECRET`, then don't ask the user a question and just fill out the entry in the secret section using the default value. If not, the behavior is the same as before.
- Throughout this, the Configuration Manager keeps track of what the secret sections are. After the `configuration.ini` is created, the Configuration Manager then adds comments above secret sections to show they should not be touched.
- Tests are also updated to test that values in secret sections are not affected by user input (i.e. they remain as the "default" values)

This is what secret sections would look like now:
```
; DO NOT MODIFY THIS SECTION
[DB_CLIENT]
user = datashark
db = datasharkdb
host = datasharkdatabase.cwnzqu4zi2kl.us-west-1.rds.amazonaws.com
port = 5432
table_name = category_labels
max_tries = 3
wait_time = 10
```

### DIFF
--- a/core/configuration.py
+++ b/core/configuration.py
@@ -12,13 +12,13 @@ class ConfigurationManager(object):
     """
     Configuration Manager
 
-    Singleton class that manages the configuration of all other modules.
+    Class that manages the configuration of all other modules.
 
     """
 
     def __init__(self):
         """
-        Virtually private constructor.
+        Initializes Configuration Manager, config not set up yet
         """
         self.question_format = "{question} [default = {default}] "
         self._config = None
@@ -31,25 +31,27 @@ class ConfigurationManager(object):
         self._config = None
         return self
 
-    def bootstrap(self, config_filepath="core/configuration.ini", input_function=None):
+    def bootstrap(self, config_filepath="core/configuration.ini", 
+                    question_filepath="core/questions.csv", input_function=None):
         """
         Do nothing if config instance already exists. Otherwise, check if
-        core/configuration.ini exists. If not, call run_setup_mode. Then load
+        core/configuration.ini exists. If not, call _run_setup_mode. Then load
         the config instance from configuration.ini
         """
         if self._config:
             return False
         self.input_function = input_function if input_function else input
         self.config_filepath = config_filepath
+        self.question_filepath = question_filepath
         if not os.path.isfile(self.config_filepath):
-            self.run_setup_mode(self.config_filepath, self.input_function)
+            self._run_setup_mode(self.config_filepath, self.question_filepath, self.input_function)
         else:
             self._config = configparser.ConfigParser()
             self._config.read(self.config_filepath)
         return True
 
 
-    def run_setup_mode(self, config_filepath, user_input):
+    def _run_setup_mode(self, config_filepath, question_filepath, user_input):
         """
         Create configuration.ini with the user_input function. In testing,
         this is a deterministic lambda function.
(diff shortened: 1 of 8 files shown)


core/configuration.py
```diff
@@ -6,73 +6,91 @@
...
+                    format='[ConfigurationManager] %(asctime)s %(levelname)s %(message)s')
 
 class ConfigurationManager(object):
-	"""
-	Configuration Manager
+    """
+    Configuration Manager
 
-	Singleton class that manages the configuration of all other modules.
+    Singleton class that manages the configuration of all other modules.
```
georgymh: Can we remove the reference to "Singleton" here since it's not really accurate?

core/configuration.py
```diff
@@ -6,73 +6,91 @@
...
 
-	def __init__(self):
-		"""
-		Virtually private constructor.
-		"""
-		self.question_format = "{question} [default = {default}] "
-		self._config = None
+    def __init__(self):
+        """
+        Virtually private constructor.
```
georgymh: Can we also remove the "virtually private constructor" reference?

core/configuration.py
```diff
@@ -6,73 +6,91 @@
...
-			return False
-		self.input_function = input_function if input_function else input
-		self.config_filepath = config_filepath
-		if not os.path.isfile(self.config_filepath):
-			self.run_setup_mode(self.config_filepath, self.input_function)
-		else:
-			self._config = configparser.ConfigParser()
-			self._config.read(self.config_filepath)
-		return True
+    def bootstrap(self, config_filepath="core/configuration.ini", input_function=None):
```
georgymh: So I noticed the tests don't test a `questions.csv` with more than one secret subsection (other than DBClient). Looking at the code it *seems* like this would work, but it would be better if our tests actually cover this.

We can make this work by passing an argument `questions_filepath` to `bootstrap()`, passing it all the way to `_run_setup_mode()`, and using a testing .csv with 2+ sections on the tests. (This also removes the magic string in the `_run_setup_mode()` method.)

core/configuration.py
```diff
@@ -6,73 +6,91 @@
...
-			question, section, key, default = row['question'], row['section'], row['key'], row['default']
-			answer = user_input(self.question_format.format(question=question, default=default))
-			answer = answer if answer else default
-			if not section in config:
-				config[section] = {}
-			config[section][key] = answer
-		with open(self.config_filepath, 'w') as configfile:
-			config.write(configfile)
-		self._config = config
+    def run_setup_mode(self, config_filepath, user_input):
```
georgymh: Could we make this method private for clarity?

neeleshdodda44: Done.

neeleshdodda44:
@nzoghb This PR explains how secret configurations work. Just FYI since all configurations should now be in `configuration.ini`.

PR #23, opened 2018-10-06
Add DMLResult and incorporate it to the Runner->Scheduler error flow

georgymh:
This PR introduces a DMLResult class. DMLResult objects are created inside the Runners once a DMLJob is executed. A DMLResult contains metadata of the job (status, error message, job type) as well as the results of the job.

This PR also makes the Runners return the DMLResults to the Scheduler, which ignores them if the jobs were successful or does retry-logic if the jobs failed. As a side note, the Scheduler's retry-logic seemed to have a bug and as an attempt to fix it I broke up the `current_jobs` property into two lists: one for DMLJob objects and another for asynchronous DMLResult objects (asynchronous because these are promise objects from Python's pool library).

The next PR will incorporate the DMLResult class to the Communication Manager for the success flow of a job.

### DIFF
--- a/core/utils/dmlresult.py
+++ b/core/utils/dmlresult.py
@@ -1,0 +1,68 @@
+from core.utils.dmljob import serialize_job, deserialize_job
+
+class DMLResult(object):
+    """
+    DML Result
+
+    This class is created by the Runners whenever they're done running a job,
+    and it's passed to the Communication Manager on success or back to the
+    Scheduler on error.
+
+    """
+
+    def __init__(
+        self,
+        status,
+        job,
+        results={},
+        error_message="",
+    ):
+        """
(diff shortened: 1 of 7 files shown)


core/utils/dmlresult.py
```diff
@@ -0,0 +1,71 @@
...
+        status,
+        job_type,
+        results={},
+        error_message="",
+    ):
+        """
+        Initializes a DML Result object.
+
+        Args:
+            status (str): status of the job [successful|failed].
```
neeleshdodda44: Nitpick: if status is only going to be `successful` or `failed`, why not just use a boolean?

kiddyboots216: more statuses in future

PR #25, opened 2018-10-15
Adding Blockchain Gateway

kiddyboots216:
Blockchain_Gateway implements core functionality (getting and setting key/value pairs from blockchain) in the API section
In the Developer section, we include broadcasting functionality to initiate and terminate decentralized training
In the Provider section, we include listening loop functionality to listen for new decentralized training and new weight updates from the blockchain
Only the setting/getting is in pytests currently, but we have tested more integrated functionality with the Communication Manager; that's not tested here because the PR for the Optimize branch still needs to be filed.


core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+            `{
+                type: ... ,
+                payload: ...
+            }`
+        """
+        pass
+
+    def update_state(self, new_state_wrapper: list) -> None:
+        """
+        Given the freshly-downloaded state, call a handler on each transaction
```
georgymh: The docs should explain or at least mention what the `new_state_wrapper` parameter is.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        that was not already present in our own state
+        """
+        new_state = dict(new_state_wrapper).get('messages')
```
georgymh: What happens if `messages` is not found in the dict?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        len_state = len(self.state)
+        for i in new_state[len_state:]:
```
georgymh: `for i ...` is commonly used for integers and indexes. Could you do `for e ...`, `for elem ...` or `for pair ...` instead for clarity?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+            self.state.append(i)
+
+    def setter(self, key: str, value: object, flag: bool = False) -> str:
+        """
+        Provided a key and a JSON/np.array object, upload the object to
+        IPFS and then store the hash as the value on the blockchain. The key
+        should be a backward reference to a prior tx
```
georgymh: What's a tx and what do you mean with "backward reference"? Can we be more clear in the docs?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        on_chain_value = self._upload(value) if value else None
+        key = on_chain_value if flag else key
+        tx = {'key': key, 'content': on_chain_value}
```
georgymh: Why do we want this to be `{'key': None, 'content': None}` if value is null? If this is intended, could we document it to avoid any surprises?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        ipfs_hash = self._content_to_ipfs(obj)
+        return str(ipfs_hash)
+
+    def _download(self, key: str) -> object:
+        """
+        Provided an on-chain key, retrieve the value from local state and
+        retrieve the Python object from IPFS
+        TODO: implement a better way to parse through state list
+        """
+        relevant_txs = [self._ipfs_to_content(tx.get('content'))
```
georgymh: This is traversing the list from index 0 to N. Why not from N to 0? (I'm assuming things are appended to the list instead of prepended.)

Also, what if the 2 `.get()`'s fail?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+    def _content_to_ipfs(self, content: dict) -> str:
+        """
+        Helper function to deploy a Python object onto IPFS, 
+        returns an IPFS hash
+        """
+        ipfs_hash = self.client.add_json(content)
+        return ipfs_hash
+
+    ##########################################################################
+    ###                         DEVELOPER SECTION                          ###
```
georgymh: Please remove the developer section from the provider codebase.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+    def handle_decentralized_learning_owner(self, model_config: object) -> None:
+        """
+        Return weights after training terminates
+        TODO: add condition to check if training for specific model terminated
+        """
+        final_weights = self.getter(header)
+        return final_weights
+
+    ##########################################################################
+    ###                          PROVIDER SECTION                          ###
```
georgymh: Everything below this section (and above if I missed it) should use the same new python syntax as the sections above. Let's make the class consistent in style.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def handle_decentralized_learning_trainer(self, tx: dict) -> None:
+        """
+        This callback will be triggered by the Listener if it finds the 
+        method signature it's looking for.
+        The parameters (model weights, model config) will be downloaded 
+        and put into the optimizer initially. So the optimizer knows this info.
+        """
+        logging.info("handling decentralized learning... {}".format(tx))
+        key = tx.get('key')
```
georgymh: Please check for errors while checking for these keys.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        intended for its node ID. Downloads the weights, and gives them to the
+        comm. mgr which gives them to the relevant optimizer 
+        -which should do the moving average.
+        """
+        logging.info("handling new weights...{}".format(tx))
+        key = tx.get('key')
+        value = tx.get('content')
+        args = {'key': key, 'content': self._ipfs_to_content(value)}
+        # TODO: Put into in-memory datastore.   
+        self.communication_manager.inform("new_weights", args)
```
georgymh: "new_weights" should be uppercase for consistency. However, ideally you should use the enum here, as well as in all other places instead of the magic strings.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        Polls blockchain for node ID in decentralized_learning() method
+        signature decentralized_learning(...<node_ids>...) method signature
+        will be the key on the blockchain; listener should look for this, and
+        if the method signature contains its node id, it will trigger a
+        callback
+        """
+        def filter(tx):
+            logging.info("tx: {}".format(tx))
+            return tx.get('key') == tx.get('content')
```
georgymh: Same as above.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        return self.filter_set(filter,
+                                self.handle_decentralized_learning_trainer)
+
+    def broadcast_new_weights(self, tx: dict) -> str:
+        """
+        broadcast_new_weights() method with all relevant parameters
+        should populate the key of new_weights with all of the nodes for which
+        these new weights are relevant. value should be IPFS hash.
+        """
+        content = tx.get('content', None)
```
georgymh: Why default to None if it will fail on the next line? We need better error handling if possible.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        key = self.client.add_json(content['weights'])
```
georgymh: Why isn't this just `_content_to_ipfs(content['weights'])`?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+    """
+    Blockchain Gateway 
+
+    The blockchain client exposes `setter` and `getter` in order to interact
+    with the blockchain.
+
+    In order for this to work, the following must be running:
+        IPFS Daemon: `ipfs daemon`
+        The lotion app: `node app_trivial.js` from the application root directory
+
```
georgymh: There's nothing explaining how listening works/is operated in this class.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        node is done training a particular model, to which the Optimizer could
+        decide it's time to communicate the new weights to the network.
+        If the Optimizer says yes, then the Communication Manager should
+        relay this info to the Listener, and the Listener uploads weights.
+        """
+        self._parse_and_run_callback(event_type, payload)
+
+    def _parse_and_run_callback(self, event_type, payload):
+        """
+        Parses an actionable_event and runs the
```
georgymh: Shouldn't it be "Parses a message_event..."?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        Polls blockchain for node ID in new_weights() method signature
+        new_weights(...<node_ids>...) method signature will be the key on the
+        blockchain; listener should look for this, and if the method signature
+        contains its node id, it will trigger a callback
+        """
+        logging.info("I'm listening for new weights!")
+        def weights_filter(tx):
+            logging.info("filtering for new weights: {}".format(tx))
+            return tx.get('key') != tx.get('content')
```
georgymh: Same as above.

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+class BlockchainGateway(object):
+    """
+    Blockchain Gateway 
+
+    The blockchain client exposes `setter` and `getter` in order to interact
+    with the blockchain.
+
+    In order for this to work, the following must be running:
+        IPFS Daemon: `ipfs daemon`
+        The lotion app: `node app_trivial.js` from the application root directory
```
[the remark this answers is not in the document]
kiddyboots216: nicolas's 'non_trivial_chain' directory

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        TODO: Refactor dependencies
+        TODO: deal with config
+        """
+        # TODO: `communication_manager` is only used in a subset of methods,
+        # consider separating
+        self.communication_manager = communication_manager
+        # TODO: We will do this once the Communication Manager is done.
+        # self.communication_manager.configure_listener(self)
+
+        config = config_manager.get_config() if config_manager else {}
```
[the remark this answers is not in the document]
kiddyboots216: existed for testing, will update

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        Do nothing.
+        """
+        pass
+
+    def get_global_state(self):
+        """
+        Gets the global state which should be a list of dictionaries
+        """
+        timeout = time.time() + 5
```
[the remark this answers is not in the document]
kiddyboots216: @nzoghb move timeout to be instance variable initialized in constructor from config

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+                continue
+        logging.info("global state:{}".format(tx_receipt.json()))
+        return tx_receipt.json()
+
+    def get_diffs(self, old_state: list, new_state: dict) -> list:
+        """
+        Iterate through oldState and newState to see any differences
+        Take action based on the differences
+        """
+        tx_diffs = [tx for tx in new_state.get('messages') if (
```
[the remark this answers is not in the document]
kiddyboots216: there is no way new_state will not have 'messages'

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+                        tx not in old_state)]
+        return tx_diffs
+
+    def get_state_diffs(self, event_filter: object) -> list:
```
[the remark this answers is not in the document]
kiddyboots216: 1- will rename
2- will update

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        new_state = self.get_global_state()
+        if new_state:
+            state_diffs = self.get_diffs(self.state, new_state)
+            filtered_diffs = [tx for tx in state_diffs if event_filter(tx)]
+            logging.info("filtered diffs:{}".format(filtered_diffs))
+            self.update_state(new_state)
+            return filtered_diffs
+
+    def check_malformed_content(self) -> None:
```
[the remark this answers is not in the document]
kiddyboots216: will remove

core/blockchain/blockchain_gateway.py
kiddyboots216: @nzoghb i have no clue what new_state_wrapper is

core/blockchain/blockchain_gateway.py
kiddyboots216: this should never happen

core/blockchain/blockchain_gateway.py
kiddyboots216: both of these are elaborated on in Quip; function docstrings are not the place to explain blockchain state design semantics

core/blockchain/blockchain_gateway.py
kiddyboots216: this is a Terminate command, purely used for testing

core/blockchain/blockchain_gateway.py
kiddyboots216: can you elaborate on 'why not from N to 0?'
also .get()'s shouldn't fail

core/blockchain/blockchain_gateway.py
kiddyboots216: need to refactor but we want feedback on this in this PR rather than proposing this as separate PR, getting changes, having to update this, etc

core/blockchain/blockchain_gateway.py
kiddyboots216: there is no way a tx will not have 'key' because of how it is constructed

core/blockchain/blockchain_gateway.py
kiddyboots216: wasn't sure whether we were using enum or strings actually, will update

core/blockchain/blockchain_gateway.py
kiddyboots216: same as above

core/blockchain/blockchain_gateway.py
kiddyboots216: think it errored when we tried to do that but let me try and revert

core/blockchain/blockchain_gateway.py
kiddyboots216: same as above? what is same as above

core/blockchain/blockchain_gateway.py
kiddyboots216: yep i copy-pasted whoops will update docstring

core/utils/event_types.py
```diff
@@ -0,0 +1,52 @@
...
+
+class MessageEventTypes(Enum):
+    """
+    Dictionary of events that the Blockchain Gateway might receive from the 
+    Communication Manager.
+
+    The event indicates what the Communication Manager wants the Gateway to
+    do with the information it is asking it to communicate.
+    """
+    NEW_WEIGHTS = "new_weights"
```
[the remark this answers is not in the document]
kiddyboots216: yes

tests/test_blockchain_gateway.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def test_blockchain_gateway_can_be_initialized(blockchain_gateway):
+    assert blockchain_gateway is not None
+
+def test_blockchain_gateway_public_interface_works(blockchain_gateway):
+    blockchain_gateway.setter('hello', 'world')
+    get_val = blockchain_gateway.getter('hello')
+    assert get_val == ['world']
+
```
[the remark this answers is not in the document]
kiddyboots216: when you set with the same key again the test fails
& sure @nzoghb

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        Gets the global state which should be a list of dictionaries
+        """
+        timeout = time.time() + 5
+        tx_receipt = None
+        while time.time() < timeout:
+            try:
+                tx_receipt = requests.get(
+                    "http://localhost:{0}/state".format(self.port))
+                tx_receipt.raise_for_status()
+                if tx_receipt:
```
neeleshdodda44: To clarify, is line 80 checking that `tx_receipt` is not `None`? If so, then wouldn't line 79 error if `tx_receipt` was `None`?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        len_state = len(self.state)
+        for i in new_state[len_state:]:
+            self.state.append(i)
+
+    def setter(self, key: str, value: object, flag: bool = False) -> str:
+        """
+        Provided a key and a JSON/np.array object, upload the object to
+        IPFS and then store the hash as the value on the blockchain. The key
+        should be a backward reference to a prior tx
+        """
```
neeleshdodda44: Nitpick: can you also add logging for this function?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def getter(self, key: str) -> list:
+        """
+        Next, provided a key, get the IPFS hash
+        from the blockchain and download the object from IPFS
+        """
+        logging.info("Getting from blockchain...")
+        self.update_state(self.get_global_state())
+        retval = self._download(key)
+        return retval
```
neeleshdodda44: Nitpick: could just do `return self._download(key)`, doesn't reduce readability IMO

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        logging.info("global state:{}".format(tx_receipt.json()))
+        return tx_receipt.json()
+
+    def get_diffs(self, old_state: list, new_state: dict) -> list:
+        """
+        Iterate through oldState and newState to see any differences
+        Take action based on the differences
+        """
+        tx_diffs = [tx for tx in new_state.get('messages') if (
+                        tx not in old_state)]
```
neeleshdodda44: Seems like you want to get all the elements in `new_state.get('messages')` that are not in `old_state`. Nitpick, but I would suggest using's `numpy`'s `setdiff1d`, i.e.
```
tx_diffs = np.setdiff1d(new_state.get('messages'), old_state)
```
`numpy` is vectorized, so if the size of the lists are large enough, the speedup may be signficant.

kiddyboots216: lists are gonna be like 1mn+ long

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        return tx_diffs
+
+    def get_state_diffs(self, event_filter: object) -> list:
+        """
+        Gets state, then finds diffs, then sets state of blockchain.
+        """
+        new_state = self.get_global_state()
+        if new_state:
+            state_diffs = self.get_diffs(self.state, new_state)
+            filtered_diffs = [tx for tx in state_diffs if event_filter(tx)]
```
neeleshdodda44: Nitpick: `filter` functions are faster than list comprehensions, so if `state_diffs` is large enough, I'd recommend using `filter`.

kiddyboots216: wtf is a filter

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
nzoghb: messages is part of the blockchain's state, I can make a note to test it on the final blockchain product (not the gateway) once we settle on lotion vs cosmos

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
nzoghb: there will be a need of this type of checking sometime in the future, removing for now

core/blockchain/blockchain_gateway.py
nzoghb: Lotion's state has to come in a superficial dict. Generally state can have other keys besides `messages` but we don't need anything additional

core/blockchain/blockchain_gateway.py
nzoghb: used `tx`

core/blockchain/blockchain_gateway.py
nzoghb: I'm resolving for now. for future reference tx is the common shorthand for transaction

core/blockchain/blockchain_gateway.py
nzoghb: Assuming this will go into the MVP
@kiddyboots216 resolve if true, if it'll be different for the MVP then comment what

core/blockchain/blockchain_gateway.py
nzoghb: @georgymh would `assert` statements here help with the gets?

core/blockchain/blockchain_gateway.py
nzoghb: @georgymh would `assert` statements help here?

core/blockchain/blockchain_gateway.py
nzoghb: it should work the same
@kiddyboots216 any updates?

core/blockchain/blockchain_gateway.py
```diff
@@ -0,0 +1,355 @@
...
+        Do nothing.
+        """
+        pass
+
+    def get_global_state(self):
+        """
+        Gets the global state which should be a list of dictionaries
+        """
+        timeout = time.time() + 5
+        tx_receipt = None
```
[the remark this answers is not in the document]
nzoghb: The instantiation sometimes fails which is why we need to do it outside of the loop, but yeah setting it to `None` won't help with the return statement

core/blockchain/blockchain_gateway.py
nzoghb: yes line 80 is redundant. but 79 would not error because `try` would catch `requests.exceptions.ConnectionError`

tests/test_blockchain_gateway.py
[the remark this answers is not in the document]
nzoghb: @kiddyboots216 not necessarily, it just appends more values to it
anyway added more tests; these should work as long as the blockchain is in a clean state before the tests are run

core/blockchain/blockchain_gateway.py
nzoghb: @georgymh lmk if you can think of a better name

core/blockchain/blockchain_gateway.py
nzoghb: @georgymh this came up above; what's the best move here? `assert` statement ok?

core/blockchain/blockchain_gateway.py
kiddyboots216: @nzoghb didn't try, you should try

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
georgymh: Where can I find this?

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
georgymh: Do you guys mind explaining why there's no way "new_state will not have 'messages'"?
Might be good to add the explanation to the docstring so other devs can auto-answer this themselves in the future.

core/utils/event_types.py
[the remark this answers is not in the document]
kiddyboots216: @nzoghb to clarify, i didn't do this bc letting you handle this PR

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
kiddyboots216: https://github.com/nzoghb/Non-Trivial-Chain/blob/master/on-chain-component/app.js

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
kiddyboots216: @nzoghb to clarify, didn't do the above ^ but you should

core/blockchain/blockchain_gateway.py
kiddyboots216: @nzoghb to clarify, didn't do above, but you should

core/blockchain/blockchain_gateway.py
kiddyboots216: @nzoghb same as above

core/blockchain/blockchain_gateway.py
kiddyboots216: @nzoghb after you commit, I'll update this to use the callback_handler convention that's going to be on master

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
georgymh: How/when are we planning to integrate the two codebases? As of now, this comment is confusing to anyone who reads this. We should make it more clear.

core/blockchain/blockchain_gateway.py
georgymh: Yes, an assert would do the trick.

core/blockchain/blockchain_gateway.py
georgymh: Ok, just elaborate on this on the docstring.

core/blockchain/blockchain_gateway.py
georgymh: Yes @nzoghb 

@kiddyboots216, what I mean is why are we traversing from the beginning of time instead of the end of time backwards?

core/blockchain/blockchain_gateway.py
georgymh: Don't assume you will use the same rest of the code on the other codebase. The modules on the other codebase can have completely different requirements from this codebase, and it's out of the scope of this codebase to review this section, so please remove.

core/blockchain/blockchain_gateway.py
georgymh: Yes

core/blockchain/blockchain_gateway.py
georgymh: asserts would work (but also please note that if these functions are only used by other functions that already do the assert, you don't need to do it on every function -- just add the precondition on the docstring)

core/blockchain/blockchain_gateway.py
georgymh: Also, please add docstrings to all methods.

core/blockchain/blockchain_gateway.py
georgymh: The .get() failing. Assert works but be aware of where this will be used -- the assert could have been done already, in which case you just need to comment that in the docstring.

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
georgymh: Why/how can the instantiation sometimes fail?

core/blockchain/blockchain_gateway.py
nzoghb: I think the way to go here would be to get `{}` if you search a key that may not be in the chain. You can't stop people from querying non-existant values

core/blockchain/blockchain_gateway.py
nzoghb: works fine

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
nzoghb: requests are async

core/blockchain/blockchain_gateway.py
nzoghb: How in depth should we have docstrings be? they're already adding a lot of bulk to an otherwise lean file

core/blockchain/blockchain_gateway.py
nzoghb: @georgymh where do we move it to?

core/blockchain/blockchain_gateway.py
nzoghb: Actually upon messing with it a little, I think the cleaner option is to have it return `{}` instead of asserting fail. Users should be able to ask for a key that doesn't exist

core/blockchain/blockchain_gateway.py
```diff
@@ -1,15 +1,9 @@
...
-import os
-import requests
-import time
 
 import ipfsapi
 
-from core.configuration import ConfigurationManager
-from core.utils.event_types import MessageEventTypes
+from core.blockchain.blockchain_utils import *
+from core.utils.event_types import *
```
neeleshdodda44: I would highly recommend not using `import *`. The person working on this file does not immediately know what is being imported, which could mean that: 

1. One might accidentally rename an imported method
2. Given a class or method being used in this file, it is not obvious whether it comes from this file or some other file (since imported classes/methods aren't explicit in the import statement).

core/blockchain/blockchain_utils.py
```diff
@@ -0,0 +1,161 @@
...
+###                                 DIFFS                                  ###
+##############################################################################
+
+def get_diffs(global_state: list, local_state: list) -> list:
+    """
+    Return list of transactions that are present in `global_state` but not in
+    `local_state`
+    """
+    len_local_state = len(local_state)
+    return global_state[len_local_state:]
```
neeleshdodda44: Nitpick: `len(local_state)` is self-documenting, you can honestly just replace with: `return global_state[len(local_state):]`

core/blockchain/blockchain_utils.py
```diff
@@ -0,0 +1,161 @@
+import json
+import logging
+import requests
+import time
+from typing import Callable
+
+from core.utils.tx_utils import TxEnum, Transaction
+
+
+# TODO: does this need to be here?
```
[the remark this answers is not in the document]
georgymh: Yes, for now follow Neel's suggestion. We will figure out how to fix logging in general in the codebase later.

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
georgymh: Bump.

kiddyboots216:
@georgymh did we decide on having everything that needs dependency injection get configured in .configure or in init?

core/blockchain/blockchain_gateway.py
georgymh: It's okay to add more lines if they help with readability of the code. Explain enough s.t. a new developer in the team doesn't need to dig into the code to know what this is doing.

georgymh:
@kiddyboots216: In configure().

core/blockchain/blockchain_gateway.py
georgymh: Remove it from this PR if it's not being used in the Service, and add it in a PR on the Notebook repo if applicable.

core/blockchain/blockchain_gateway.py
```diff
@@ -1,15 +1,9 @@
...
 
 
 logging.basicConfig(level=logging.DEBUG,
```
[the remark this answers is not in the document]
georgymh: Note that you will have to import logging once you remove the import stars up there.

core/blockchain/blockchain_utils.py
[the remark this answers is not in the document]
kiddyboots216: addressed

core/blockchain/blockchain_gateway.py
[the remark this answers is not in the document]
kiddyboots216: addressed

core/blockchain/blockchain_gateway.py
```diff
@@ -52,20 +61,20 @@ def broadcast_decentralized_learning(self, model_config: object)-> str:
...
         TODO: check if training even started
         """
-        tx_receipt = setter(key, None, self.client, self.port)
+        tx_receipt = setter(self.client, key, self.port, None)
         return tx_receipt
 
     def handle_decentralized_learning_owner(self, model_config: object) -> None:
         """
-        Return weights after training terminates
+        Return weights after training terminates2
```
neeleshdodda44: Nitpick af but was this change accidental?

core/blockchain/blockchain_gateway.py
```diff
@@ -151,7 +162,6 @@ def listen_decentralized_learning(self) -> list:
         callback
         """
         def filter(tx: dict) -> str:
-            logging.info("tx: {}".format(tx))
```
neeleshdodda44: I presume you'll add these logging lines back in when you're done testing, right?

core/blockchain/blockchain_gateway.py
```diff
@@ -165,7 +175,6 @@ def listen_new_weights(self) -> None:
         """
         logging.info("I'm listening for new weights!")
         def weights_filter(tx):
-            logging.info("filtering for new weights: {}".format(tx))
```
neeleshdodda44: Same as above.

core/blockchain/blockchain_gateway.py
```diff
@@ -174,7 +183,6 @@ def listen_terminate(self) -> None:
         Polls blockchain to see whether to terminate
         """
         def filter(tx):
-            logging.info("tx: {}".format(tx))
```
neeleshdodda44: Same as above.

core/blockchain/blockchain_utils.py
```diff
@@ -112,13 +111,12 @@ def get_global_state(host: str, port: int, timeout: int) -> object:
     tx_receipt = None
     while time.time() < timeout:
         try:
-            retval = make_getter_call(host, port).json()
+            tx_receipt = make_getter_call(host, port).json()
             break
         except (UnboundLocalError, requests.exceptions.ConnectionError) as e:
             logging.info("HTTP GET error, got: {0}".format(e))
             continue
-    logging.info("global state: {}".format(retval))
```
neeleshdodda44: Same as above.

core/blockchain/blockchain_utils.py
```diff
@@ -127,7 +125,7 @@ def getter(client: object, key: str, local_state: list, port: int, timeout: int,
     object from IPFS. This pulls from the global state but DOES NOT update the
     local state
     """
-    logging.info("Getting from blockchain...")
```
neeleshdodda44: Same as above.

core/blockchain/tx_utils.py
```diff
@@ -16,9 +16,12 @@ class Transaction(object):
...
-    def __init__(self, key: str, value: object) -> None:
+    def __init__(self, key: str, value: str) -> None:
         self.key = key
         self.value = value
 
-    def __repr__(self) -> dict:
+    def get_tx(self) -> dict:
+        # print(self.key)
+        # print(self.value)
+        # print('jjj', {TxEnum.KEY.name: [self.key], TxEnum.CONTENT.name: self.value})
```
neeleshdodda44: I presume you were using these print statements to test? Please delete when you're done testing.

tests/test_blockchain_gateway.py
```diff
@@ -0,0 +1,61 @@
+import tests.context	
+import pytest	
+from core.configuration import ConfigurationManager	
+from core.blockchain.blockchain_gateway import BlockchainGateway
+from core.utils.enums import RawEventTypes
+from core.blockchain.tx_utils import TxEnum
+@pytest.fixture
```
neeleshdodda44: Please add two empty lines above this line.

tests/test_blockchain_gateway.py
```diff
@@ -0,0 +1,61 @@
...
+#     pass	
+# def test_handle_new_weights(blockchain_gateway):	
+#     """To be implemented."""	
+#     pass	
+# def test_listen_terminate(blockchain_gateway):	
+#     """To be implemented."""	
+#     pass	
+# def test_handle_terminate(blockchain_gateway):	
+#     """To be implemented."""	
+#     pass
```
neeleshdodda44: Will review these tests in detail when the tests pass.

tests/test_blockchain_utils.py
```diff
@@ -23,44 +23,58 @@ def ipfs_client(config):
...
+    get_val_after = getter(
         ipfs_client,
-        config.get('BLOCKCHAIN', 'test_key'),
+        config.get('BLOCKCHAIN', 'test_multiple_key'),
         [],
         config.getint('BLOCKCHAIN', 'http_port'),
         config.getint('BLOCKCHAIN', 'timeout')
         )
-    assert get_val == ['World!', 'World!', 'World!']
+    assert get_val_after == get_val_before + ["'World!'", "'World!'"]
```
neeleshdodda44: Same as above.

tests/test_blockchain_gateway.py
kiddyboots216: done

core/blockchain/blockchain_gateway.py
kiddyboots216: It's being used in the Service for internal testing. It will be removed as soon as the Notebook can test the Service.

kiddyboots216:
Removed Developer section.

PR #26, opened 2018-10-20
Communication_Manager & Optimizer

kiddyboots216:
This is a big one.
Tons of changes.
@georgymh rewrote communication_manager and fedavgoptimizer pretty much entirely
Updated Scheduler and Runner for DMLResults flow
Communication and Averaging jobs are currently nonfunctional; I'm going to make separate PRs for those
The biggest test is the first one that runs in tests/test_communication_manager because...well, just take a look. I basically can't unit-test, not sure why, initializing the model will hang.


core/communication_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+        Creates a new session and optimizer based on the parameters sent by a
+        model developer. It then asks the optimizer what's the first thing to do
+        and the service starts working on that session from that point on.
+        """
+        # NOTE: We removed the 'optimizer_type' argument since for the MVP we're
+        # only considering the 'FederatedAveragingOptimizer' for now.
+        # TODO: We need to incorporate session id's when we're ready.
+        logging.info("New optimizer session is being set up...{}".format(payload))
+        # session_id = payload.get('key')
```
georgymh: I thought the 'key' was a blockchain-specific value, not the `session_id`?

@nzoghb @kiddyboots216 (Can we write a quick & dirty docs about this?)

kiddyboots216: It'll get fixed when we merge in Blockchain Gateway and do integration testing, I'm sure.
Either way doesn't matter because it's in a comment.

georgymh: Ok. Will remove the commented code for now since it might be inaccurate.

core/communication_manager.py
```diff
@@ -0,0 +1,155 @@
...
+        event_type, payload = self.optimizer.kickoff()
+        callback = callback_handler(
+            event_type,
+            self.ACTIONABLE_EVENT_TYPE_2_CALLBACK,
+            ActionableEventTypes.NOTHING.value
+        )
+        callback(payload)
+        logging.info("Kickoff complete.")
+
+    def _transform_and_split(self, payload):
```
neeleshdodda44: If this hasn't been implemented yet, can you remove this method or comment it out?

core/communication_manager.py
```diff
@@ -0,0 +1,155 @@
...
+        logging.info("Transforming and splitting... (NOT IMPLEMENTED YET)")
+        pass
+
+    def _schedule_job(self, dmljob_obj):
+        """
+        Helper function that schedules a DML Job through the Scheduler.
+        """
+        if not self.scheduler:
+            raise Exception("Communication Manager needs to be configured first!")
+        if not dmljob_obj or not isinstance(dmljob_obj, DMLJob):
```
neeleshdodda44: Redundant, can just do `not isinstance(dmljob_obj, DMLJob)`

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
+from core.utils.event_types import ActionableEventTypes, RawEventTypes, callback_handler
+from core.utils.dmljob import deserialize_job
+from core.utils.dmlresult import DMLResult
+
```
neeleshdodda44: Can you add an extra line here? Our standard is two empty lines between import statements and rest of code.

core/utils/dmljob.py
```diff
@@ -41,6 +41,9 @@ def __init__(
         self.hyperparams = hyperparams
         self.label_column_name = label_column_name
 
+    def add_weights(self, serialized_weights):
```
georgymh: I'm assuming `serialized_weights` is the output of calling `core.utils.keras.serialize_weights()` on a `job.weights`. Can you confirm?

What's the general purpose of this function and why does a DMLJob need to have both `.job` and `.serialized_weights`?

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		Case 2: Termination
+		Case 3: Init ???
+		"""
+		# event_type = payload['info_type']
+		# callback = callback_handler(event_type, self.LEVEL_2_INFO_CALLBACKS)
+		# NOTE: Hardcoded to handle new weights for now...
+		callback = callback_handler("new_weights", self.LEVEL_2_INFO_CALLBACKS)
+		return callback(payload)
+
+	def _update_weights(self, weights):
```
neeleshdodda44: Nitpick but this helper function seems to just be a one liner. I see that it is only called in `handle_job_done`, so why not just remove this function and replace `self._update_weights(new_weights)` with `self.job.weights = new_weights` on line 50?

core/runner.py
```diff
@@ -103,23 +97,42 @@ def run_job(self, job):
...
-                    job_type=job.job_type,
+                    job=job,
                     results={
-                        'initial_weights': initial_weights,
+                        'weights': initial_weights,
                     },
                     error_message="",
                 )
+            elif job.job_type == 'average':
+                raise Exception("averaging is not yet supported!")
```
neeleshdodda44: Did you mean to still leave this here?

core/utils/dmljob.py
```diff
@@ -62,7 +65,7 @@ def serialize_job(dmljob_obj):
     return {
         'weights': serialize_weights(dmljob_obj.weights),
         'job_data': rest
-    }
+    } if dmljob_obj.weights else {'job_data': rest}
```
georgymh: When is `dmljob_obj.weights` won't be defined? `communicate` and possibly `average` type of jobs? Please confirm the intention of this check and if this is the case, we may need to create new types of (job) classes that the Scheduler and Runner can take in the upcoming PRs.

core/runner.py
```diff
@@ -103,23 +97,42 @@ def run_job(self, job):
...
+                results = DMLResult(
+                    status='successful',
+                    job=job,
+                    results={
+                        'weights': averaged_weights
+                    },
+                    error_message=""
+                )
+            elif job.job_type == 'communicate':
+                assert False, "communication is not yet supported!"
```
neeleshdodda44: Same as above.

core/runner.py
georgymh: Yes. The upcoming PR will implement this. Same with "communicate".

core/utils/event_types.py
```diff
@@ -0,0 +1,59 @@
...
+
+    This is essentially a 'memory' for the Optimizer so these are all past.
+    i.e., a "TRAIN" event would contain auxiliary information about what
+    occurred, in the form of a DMLResults object.
+
+    Contains all the functionality of what the Communication Manager might have
+    previously done and therefore is not specific to any one Optimizer.
+    """
+    NEW_SESSION = "new_session"
+    NEW_WEIGHTS = "new_weights"
```
neeleshdodda44: Nitpick but should `NEW_SESSION` and `NEW_WEIGHTS` be capitalized too?

georgymh: Will do

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+        'split': 0.004,
+    }
+    return hyperparams
+
+config_manager = ConfigurationManager()
+config_manager.bootstrap(
+    config_filepath='tests/artifacts/communication_manager/configuration.ini'
+)
+
+def test_communication_manager_can_initialize_and_train_model():
```
neeleshdodda44: If you can't split this up, can you identify which code in the test corresponds to which task it's trying to accomplish? Like a comment above a code block could be `# (1), (2), (3)

core/communication_manager.py
georgymh: True

core/communication_manager.py
georgymh: I think some things would fail if we remove, because the Listener assumes this type of job exists. It's not a biggie for now since you'll be done implementing the functionality to make this work pretty soon.

core/communication_manager.py
kiddyboots216: @georgymh then why did you add that line when what i had was fine lmao

core/fedavgoptimizer.py
kiddyboots216: @neeleshdodda44 wasn't sure since this standard isn't followed in rest of codebase

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    assert len(scheduler.processed) == 2
+    # TODO: I can't get the below AssertionError test to work since Georgy's code won't raise errors but it WORKS
+    # try:
+    #     scheduler.start_cron(period_in_mins = 0.05)
+    #     time.sleep(10)
+    #     scheduler.stop_cron()
+    #     assert len(scheduler.processed) == 2
+    #     assert False, "Runner should have been told to communicate"
+    # except AssertionError as e:
+    #     assert str(e) == "communication is not yet supported!"
```
neeleshdodda44: You're not going to be able to catch the exception, since the current functionality of the Runner catches errors and returns it in DMLResult. See line 131. If a job fails more than 3 times (see scheduler), then the job is discarded. 

So I'm not sure whether this test can be implemented as of now.

core/communication_manager.py
georgymh: I meant to remove `not dmljob_obj` and use `not isinstance(dmljob_obj, DMLJob)`

core/utils/dmljob.py
kiddyboots216: yes

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+def test_communication_manager_can_be_initialized():
+    communication_manager = CommunicationManager(config_manager)
+    assert communication_manager is not None
```
neeleshdodda44: `assert communication_manager`

core/utils/dmljob.py
kiddyboots216: no, 'initialize' jobs.

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    scheduler = DMLScheduler(config_manager, communication_manager)
+    scheduler.reset()
+    communication_manager.configure(scheduler)
+    new_session_event = {
+            "key": None, 
+            "content": {
+                "optimizer_params": {}, 
+                "serialized_job": serialize_job(make_initialize_job(make_model_json()))}}
+    communication_manager.inform(RawEventTypes.NEW_SESSION.value, new_session_event)
+    assert communication_manager.optimizer is not None
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
kiddyboots216: gotcha

core/utils/dmljob.py
georgymh: Ok.

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    # except AssertionError as e:
+    #     assert str(e) == "communication is not yet supported!"
+def test_communication_manager_can_be_initialized():
+    communication_manager = CommunicationManager(config_manager)
+    assert communication_manager is not None
+
+def test_communication_manager_creates_new_sessions():
+    """
+    Takes in initialized_communication_manager because the initialization is all that matters.
+    Ensures that upon receiving an initialization job, the commmgr will make an optimizer.
```
neeleshdodda44: Super nitpicky but throughout the documentation, I've seen `Communication Manager`, `CommMgr` and `commgr`. Can we decide on one name and keep it consistent?

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    communication_manager = CommunicationManager(config_manager)
+    scheduler = DMLScheduler(config_manager, communication_manager)
+    scheduler.reset()
+    communication_manager.configure(scheduler)
+    new_session_event = {
+            "key": None, 
+            "content": {
+                "optimizer_params": {}, 
+                "serialized_job": serialize_job(make_initialize_job(make_model_json()))}}
```
neeleshdodda44: Can you move the `}` to new lines? End result would look like:

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+def test_communication_manager_fails_if_not_configured():
+    """
+    Ensures that commmgr won't be able to function if it's not configured
+    """
+    communication_manager = CommunicationManager(config_manager)
+    new_session_event = {
+            "key": None, 
+            "content": {
+                "optimizer_params": {}, 
+                "serialized_job": serialize_job(make_initialize_job(make_model_json()))}}
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    communication_manager = CommunicationManager(config_manager)
+    scheduler = DMLScheduler(config_manager, communication_manager)
+    scheduler.reset()
+    communication_manager.configure(scheduler)
+    true_job = make_initialize_job(make_model_json())
+    new_session_event = {
+            "key": None, 
+            "content": {
+                "optimizer_params": {}, 
+                "serialized_job": serialize_job(true_job)}}
```
neeleshdodda44: Same as above.

core/communication_manager.py
```diff
@@ -2,6 +2,7 @@
 
 from core.utils.event_types import RawEventTypes, ActionableEventTypes, callback_handler
 from core.fedavgoptimizer import FederatedAveragingOptimizer
+from core.utils.dmljob import DMLJob
```
kiddyboots216: unnecessary import

core/communication_manager.py
```diff
@@ -49,6 +48,8 @@ def __init__(self, config_manager):
         }
         logging.info("Communication Manager is set up!")
 
+    # Setup Methods
+
```
kiddyboots216: i added this in my commit which i can't push because yours has merge conflicts

core/communication_manager.py
georgymh: It's okay, it's defensive and possibly helps us track the bug to the source. Something that could happen is that we could repurpose this routine to do something else than submitting to the Scheduler, in which case having had this check here helps us identify bugs earlier.

core/communication_manager.py
kiddyboots216: line 135 is unnecessary

core/communication_manager.py
neeleshdodda44: @kiddyboots216 you kinda hijacked this thread so I'll let you resolve this one

tests/test_communication_manager.py
kiddyboots216: sadboi

core/utils/dmljob.py
kiddyboots216: - yes
- for averaging but i'll probably end up changing this in a subsequent PR for the Averaging Job

georgymh: Ok, will add a comment about that right now. We will let it slide to `master` momentarily.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		"""
+		Parses an event dictionary into a level 1 callback.
+		If the callback is not defined, it does nothing.
+		Otherwise, calls the callback.
+		"""
+		callback = callback_handler(event_type, self.LEVEL_1_CALLBACKS)
+		return callback(payload)
+
+	def do_nothing(self, payload):
+		pass
```
georgymh: Since `ask()` returns a tuple, this should return `None, None` or something similar.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		"""
+		job = self.job
+		self.listen_iterations += 1
+		if self.listen_iterations > self.listen_bound:
+			job.job_type = "train"
+			return ActionableEventTypes.SCHEDULE_JOB.value, job
+		else:
+			return ActionableEventTypes.NOTHING.value, None
+
+	def kickoff(self):
```
georgymh: Can you move public methods to the top of the file?

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		Kwargs is a dictionary containing all the info needed to initialize the Optimizer.
+		This includes any number of properties such as thresholds, etc.
+		"""
+		kwargs = content.get('optimizer_params')
+		serialized_job = content.get('serialized_job')
+		# The Optimizer knows its job
+		self.job = deserialize_job(serialized_job)
+		self.listen_iterations = kwargs.get('listen_iterations', 0)
+		self.listen_bound = kwargs.get('listen_bound', 1)
+		self.LEVEL_1_CALLBACKS = {
```
georgymh: What are level 1, 2, and 3 callbacks? Please write some docstring about them.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
+from core.utils.event_types import ActionableEventTypes, RawEventTypes, callback_handler
```
georgymh: Could you rename the file to `fed_avg_optimizer.py`?

core/runner.py
```diff
@@ -107,29 +111,31 @@ def run_job(self, job):
...
+                #     status='successful',
+                #     job=job,
+                #     results={
+                #         'weights': job.weights
+                #     },
+                #     error_message=""
+                # )
         except Exception as e:
-            logging.error(str(e))
+            logging.error("RunnerError: " + str(e))
```
neeleshdodda44: Nitpick: s/RunnerError/Runner Error:

core/utils/dmljob.py
```diff
@@ -78,11 +88,14 @@ def deserialize_job(serialized_job):
         DMLJob: A DML Job object from serialized_job.
     """
     rest = serialized_job['job_data']
+    weights = None
+    if 'weights' in serialized_job and not serialized_job['weights'] is None:
+        weights = deserialize_weights(serialized_job['weights'])
```
neeleshdodda44: Why not just do:
```
weights = deserialize_weights(serialized_job.get['weights']) if serialized_job.get('weights', None) else None
```
At the very least, I think the if statement can be simplified to `if serialized_job.get('weights', None)`

core/runner.py
georgymh: I'm following python convention (AssertionError, TypeError, etc.).

neeleshdodda44: My bad, I totally misunderstood.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+
+	"""
+	def __init__(self, content):
+		"""
+		Kwargs is a dictionary containing all the info needed to initialize the Optimizer.
+		This includes any number of properties such as thresholds, etc.
+		"""
+		kwargs = content.get('optimizer_params')
+		serialized_job = content.get('serialized_job')
+		# The Optimizer knows its job
```
georgymh: Put this comment next to the next line.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+
+class FederatedAveragingOptimizer(object):
+	"""
+	Federated Averaging Optimizer
+
+	Implements the Federated Averaging algorithm (with some tweaks)
+	and acts as a "brain" for the Unix Service
+
+	"""
+	def __init__(self, content):
```
georgymh: Can you rename "content" to something more descriptive? "content" is too general.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		"""
+		Kwargs is a dictionary containing all the info needed to initialize the Optimizer.
+		This includes any number of properties such as thresholds, etc.
+		"""
```
georgymh: Can you add start and finished logging to the init? Just like the Communication Manager does it.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+			RawEventTypes.NEW_WEIGHTS.value: self._handle_new_weights,
+			RawEventTypes.TERMINATE.value: self._done_everything,
+			RawEventTypes.NOTHING.value: self.do_nothing,
+		}
+
+	def handle_job_done(self, result):
+		"""
+		Level 1 Callback to handle the payload of a DMLResult
+		"""
+		# assert payload is a DMLResult
```
georgymh: Redundant comment. The code below it is self-documenting.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		assert isinstance(result, DMLResult)
+		# assume success for now
```
georgymh: What do you mean with this? If it's relevant, please make it clearer and consider moving it to the docstring.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		# get the DMLJob from the DMLResult
```
georgymh: This is misleading because the DMLJob is not retrieved from the DMLResult in this method. Consider removing it.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		job_type = result.job.job_type
+		callback = callback_handler(job_type, self.LEVEL_2_JOB_CALLBACKS)
+		return callback(result)
+
+	def handle_new_info(self, payload):
+		"""
+		Level 1 Callback to handle the info from the network
+		Case 1: New weights
+		Case 2: Termination
+		Case 3: Init ???
```
georgymh: Why the "???"?

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		"""
+		# event_type = payload['info_type']
```
georgymh: Can you remove the commented code out if we're not using it?

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		}
+		self.LEVEL_2_INFO_CALLBACKS = {
+			RawEventTypes.NEW_WEIGHTS.value: self._handle_new_weights,
+			RawEventTypes.TERMINATE.value: self._done_everything,
+			RawEventTypes.NOTHING.value: self.do_nothing,
+		}
+
+	def handle_job_done(self, result):
+		"""
+		Level 1 Callback to handle the payload of a DMLResult
```
georgymh: Can you add in the docstrings the side effects of the function (how it changes the state, i.e., changes the weights of the job)?

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+
+	def _handle_new_weights(self, payload):
+		"""
+		Level 2 Callback to handle new weights info
+		Updates state and then returns appropriate DMLJob
+		"""
+		job = self.job
+		job.job_type = "average"
+		serialized_weights = payload["content"]
+		job.add_weights(serialized_weights)
```
georgymh: Please add a comment here that this might change in the next PR.

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+	def _done_initializing(self, payload):
+		"""
+		Level 2 Callback to handle a DMLResult of an initialization job
+		Should return a DMLJob of type Train
+		"""
+		job = self.job
+		job.job_type = "train"
+		return ActionableEventTypes.SCHEDULE_JOB.value, job
+	
+	def _done_everything(self, payload):
```
georgymh: Please rename to something more accurate, like "_terminate_session()".

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    assert communication_manager is not None
+
+def test_communication_manager_creates_new_sessions():
+    """
+    Takes in initialized_communication_manager because the initialization is all that matters.
+    Ensures that upon receiving an initialization job, the commmgr will make an optimizer.
+    """
+    communication_manager = CommunicationManager(config_manager)
+    scheduler = DMLScheduler(config_manager, communication_manager)
+    scheduler.reset()
```
georgymh: Same as above.

tests/test_communication_manager.py
georgymh: And same as above for ease-of-read.

georgymh: And same as above for ease-of-read.

tests/test_communication_manager.py
georgymh: Same as above.

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    communication_manager = CommunicationManager(config_manager)
+    new_session_event = {
+            "key": None, 
+            "content": {
+                "optimizer_params": {}, 
+                "serialized_job": serialize_job(make_initialize_job(make_model_json()))}}
+    try:
+        communication_manager.inform(RawEventTypes.NEW_SESSION.value, new_session_event)
+        assert False, "this should have raised an exception"
```
georgymh: Can you move this to after the try-except block? And please change it to `raise Exception("...")`

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    except Exception as e:
+        assert str(e) == "Communication Manager needs to be configured first!"
+
+def test_communication_manager_can_inform_the_optimizer():
+    """
+    Ensures that commmgr can tell the optimizer of something, and that the job will transfer
+    """
+    communication_manager = CommunicationManager(config_manager)
+    scheduler = DMLScheduler(config_manager, communication_manager)
+    scheduler.reset()
```
georgymh: Same as above.

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    communication_manager.configure(scheduler)
+    true_job = make_initialize_job(make_model_json())
+    new_session_event = {
+            "key": None, 
+            "content": {
+                "optimizer_params": {}, 
+                "serialized_job": serialize_job(true_job)}}
+    communication_manager.inform(RawEventTypes.NEW_SESSION.value, new_session_event)
+    opt_job = communication_manager.optimizer.job
```
georgymh: Can we rename this to `optimizer_job`? "opt" sounds too much like optional.

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+
+from core.communication_manager import CommunicationManager
+from core.utils.event_types     import RawEventTypes, ActionableEventTypes
+from core.runner                import DMLRunner
+from core.scheduler             import DMLScheduler
+from custom.keras               import get_optimizer
+from models.keras_perceptron    import KerasPerceptron
+from core.utils.dmljob          import DMLJob, serialize_job, deserialize_job
+from core.configuration         import ConfigurationManager
+
```
georgymh: Can we add an extra line here?

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+def make_hyperparams():
+    hyperparams = {
+        'averaging_type': 'data_size',
+        'batch_size': 4,
+        'epochs': 1,
+        'split': 0.004,
+    }
+    return hyperparams
+
+config_manager = ConfigurationManager()
```
georgymh: Can we move this block to the top before the method definitions?

tests/test_communication_manager.py
```diff
@@ -0,0 +1,163 @@
...
+    scheduler.reset()
+    communication_manager.configure(scheduler)
+    new_session_event = {
+            "key": None, 
+            "content": {
+                "optimizer_params": {}, 
+                "serialized_job": serialize_job(make_initialize_job(make_model_json()))}}
+    communication_manager.inform(RawEventTypes.NEW_SESSION.value, new_session_event)
+    scheduler.start_cron(period_in_mins = 0.05)
+    time.sleep(10)
```
georgymh: We may need to do something different other than sleeping for an arbitrary chosen amount of time (10 seconds) to avoid wasting unnecessary time. What about a while loop sleeping every second and checking for the ending condition? We would need to add a timeout here though (this could be 10 or 15).

tests/test_dmlscheduler.py
```diff
@@ -11,13 +11,14 @@
 from core.utils.dmljob          import DMLJob
 from core.scheduler             import DMLScheduler
 from core.configuration         import ConfigurationManager
-
+from core.communication_manager import CommunicationManager
```
georgymh: Extra line here.

tests/test_dmlscheduler.py
```diff
@@ -11,13 +11,14 @@
...
 
 config_manager = ConfigurationManager()
 config_manager.bootstrap(
     config_filepath='tests/artifacts/runner_scheduler/configuration.ini'
 )
-scheduler = DMLScheduler(config_manager)
+communication_manager = CommunicationManager(config_manager)
+scheduler = DMLScheduler(config_manager, communication_manager)
```
georgymh: I'm curious to know how this doesn't fail if `CommunicationManager` doesn't get configured with the `scheduler`...

core/utils/dmljob.py
georgymh: Not only do we need to check if 'weights' are in `serialized_job` but **also that the weights are not `None`**. This extra check was introduced by one of my commits of making serialized DMLJob always have the `weights` key in them for consistency.

tests/test_communication_manager.py
kiddyboots216: done

core/fedavgoptimizer.py
```diff
@@ -0,0 +1,144 @@
...
+		return ActionableEventTypes.TERMINATE.value, None
+	
+	def _average_iter(self, payload):
+		"""
+		Level 2 Callback to handle a DMLResult of an averaging job
+		Should only return a DMLJob of type Train if averaging is done
+		"""
+		job = self.job
+		self.listen_iterations += 1
+		if self.listen_iterations > self.listen_bound:
```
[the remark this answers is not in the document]
kiddyboots216: we'll find out when we get there lol. most of the optimizer is untested.

core/fedavgoptimizer.py
kiddyboots216: done

core/fedavgoptimizer.py
kiddyboots216: Done

core/fedavgoptimizer.py
kiddyboots216: no bc then I will forget!

core/fedavgoptimizer.py
kiddyboots216: haven't decided on this schema yet

core/fedavgoptimizer.py
kiddyboots216: done

core/fedavgoptimizer.py
kiddyboots216: It's relevant -we're assuming that the DMLResult is 'successful' in terms of what the optimizer thinks (i.e. has enough training accuracy or smth)

core/fedavgoptimizer.py
kiddyboots216: done

core/fedavgoptimizer.py
kiddyboots216: done

core/fedavgoptimizer.py
kiddyboots216: In fact, it is the result of doing a .get("content") on what is passed in.

kiddyboots216: CommMgr does .get("content") and passes that into the optimizer. So this is, literally, content.

core/fedavgoptimizer.py
kiddyboots216: what

tests/test_communication_manager.py
kiddyboots216: done

tests/test_communication_manager.py
kiddyboots216: done

tests/test_communication_manager.py
kiddyboots216: oops

tests/test_communication_manager.py
kiddyboots216: jk i can't it all happens in one line

core/fedavgoptimizer.py
kiddyboots216: It's gonna do more in the future

georgymh:
One more thing:

- We need to test the correctness of the DMLResult returned by the Runners for each job type

core/fedavgoptimizer.py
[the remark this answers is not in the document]
georgymh: Then add a comment about this or remove it and defer the discussion for the next PR.

georgymh:
Also, running the Scheduler's cron job on the main thread may cause unexpected behavior, and we can't instantiate new threads during tests. For the Communication Manager tests, let's manually run the next job by using the proper routine in the Scheduler.

core/fedavgoptimizer.py
georgymh: Please don't resolve the conversation if you didn't initialize it, as they got lost without actually being solved.

I'm not convinced with your answer. How about renaming it to something like "initialization_payload"?

core/fedavgoptimizer.py
kiddyboots216: done

core/fedavgoptimizer.py
kiddyboots216: done

core/utils/dmljob.py
kiddyboots216: Looks good

core/fedavgoptimizer.py
kiddyboots216: No, it's named this way so that you can traceback where the argument is coming from.
The 'init' method doesn't need to have an argument with the name 'initialization' used to make it more descriptive

kiddyboots216:
I'm not able to replicate the desired behavior using scheduler.runners_run_next_job() and I'm not convinced that multiprocessing works in the Scheduler since (a) there's a decent amount of documentation suggesting that proper multiprocessing for Keras looks different than what we have and (b) There are no tests for multiprocessing, so I'd argue that the behavior we see from running the Scheduler's cron job is more of "what we expect" then just iteratively calling scheduler.runners_run_next_job() with variable-length sleep() in-between.

core/utils/federated_learning_utils.py
```diff
@@ -2,6 +2,7 @@
 
 from core.utils.keras import serialize_weights, deserialize_single_weights
 
+def unweighted
```
neeleshdodda44: What is this line doing?

core/utils/enums.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Helper function which returns a callback from a callback
+    dict when given a callback type.
+    """
+    assert isinstance(callback_dict, dict), \
+        "callback_dict was not a dict, it was: {}".format(type(callback_dict))
+    callback = callback_dict[default]
+    if callback_type in callback_dict:
+        callback = callback_dict[callback_type]
+    return callback
```
neeleshdodda44: Could just simplify to `return callback_dict.get(callback_type, callback_dict[default])`

core/fed_avg_optimizer.py
```diff
@@ -0,0 +1,157 @@
...
+		logging.info("Setting up Optimizer")
+		kwargs = content.get('optimizer_params')
+		serialized_job = content.get('serialized_job')
+		self.job = deserialize_job(serialized_job) # The Optimizer knows its job
+		self.listen_iterations = kwargs.get('listen_iterations', 0)
+		self.listen_bound = kwargs.get('listen_bound', 1)
+		self.LEVEL_1_CALLBACKS = {
+			RawEventTypes.JOB_DONE.name: self._handle_job_done,
+			RawEventTypes.NEW_INFO.name: self._handle_new_info,
+			RawEventTypes.NOTHING.name: self._do_nothing,
```
georgymh: Why would the Communication Manager ask the Optimizer what to do with "NOTHING"?

core/fed_avg_optimizer.py
```diff
@@ -0,0 +1,157 @@
...
+		}
+		self.LEVEL_2_JOB_CALLBACKS = {
+			JobTypes.JOB_TRAIN.name: self._done_training,
+			JobTypes.JOB_INIT.name: self._done_initializing,
+			JobTypes.JOB_AVG.name: self._average_increment,
+			JobTypes.NOTHING.name: self._do_nothing,
```
georgymh: The "NOTHING" job type shouldn't exist.

core/fed_avg_optimizer.py
```diff
@@ -0,0 +1,157 @@
...
+		}
+		self.LEVEL_2_INFO_CALLBACKS = {
+			RawEventTypes.NEW_WEIGHTS.name: self._handle_new_weights,
+			RawEventTypes.TERMINATE.name: self._terminate_session,
+			RawEventTypes.NOTHING.name: self._do_nothing,
```
georgymh: Is this realistic? If so, what's an example?

core/utils/enums.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def callback_handler_no_default(
```
neeleshdodda44: Nitpick but you don't need two functions to handle a no default case. Just have an extra boolean parameter called `can_default` in one function, which allows you to default or not. Just seems like a lot of code/documentation is repeated unnecessarily.

core/utils/dmljob.py
```diff
@@ -40,6 +40,17 @@ def __init__(
         self.weights = weights
         self.hyperparams = hyperparams
         self.label_column_name = label_column_name
+        self.list_of_serialized_weights = []
```
georgymh: This shouldn't be here. This seems to be for the averaging PR.

core/utils/dmljob.py
georgymh: Let's remove this since it's not being used at all. Let's table it for the next PR.

core/utils/enums.py
```diff
@@ -0,0 +1,102 @@
...
+
+    JOB_COMM: Optimizer wants Runner to push the weights onto
+    IPFS and then communicate this to rest of nodes via blockchain
+    """
+    JOB_TRAIN = "TRAIN"
+    JOB_VAL = "VALIDATE"
+    JOB_INIT = "INITIALIZE"
+    JOB_AVG = "AVERAGE"
+    JOB_COMM = "COMMUNICATE"
+    NOTHING = "NOTHING"
```
georgymh: This should be removed -- there shouldn't exist a NOTHING job type.

core/utils/enums.py
```diff
@@ -0,0 +1,102 @@
...
+
+    NEW_INFO: Sent to Communication Manager by Gateway, means that
+    the payload contains info from the blockchain
+    """
+    NEW_SESSION = "NEW_SESSION"
+    NEW_WEIGHTS = "NEW_WEIGHTS"
+    TERMINATE = "TERMINATE"
+    JOB_DONE = "JOB_DONE"
+    NEW_INFO = "NEW_INFO"
+    NOTHING = "NOTHING"
```
georgymh: Is this realistic? If so, what's an example?

core/runner.py
```diff
@@ -83,16 +82,20 @@ def _train(self, job):
...
 
         NOTE2: Assumes 'initial_weights' are the actual weights and not a path.
         """
         # Get the right dataset iterator based on the averaging type.
-        serialized_model, framework_type, initial_weights, hyperparams, labeler = job.serialized_model, job.framework_type, job.weights, job.hyperparams, job.label_column_name
+        serialized_model = job.serialized_model
+        framework_type = job.framework_type
+        initial_weights = job.weights
+        hyperparams = job.hyperparams
+        labeler = job.label_column_name
```
neeleshdodda44: Nitpick but why initialize all these variables? I think it's more readable and less lines of code to just use `job.*`

core/runner.py
```diff
@@ -145,11 +148,15 @@ def _validate(self, job, custom_iterator=None):
...
+        Returns a DMLResult with the metrics returned by the model.
 
         NOTE: Assumes 'weights' are the actual weights and not a path.
         """
-        serialized_model,framework_type,weights,hyperparams,labeler =job.serialized_model,job.framework_type,job.weights,job.hyperparams,job.label_column_name 
+        serialized_model = job.serialized_model
+        framework_type = job.framework_type
+        weights = job.weights
+        hyperparams = job.hyperparams
+        labeler = job.label_column_name
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
```diff
@@ -0,0 +1,153 @@
...
+            "serialized_job": serialized_job
+        }
+    }
+    try:
+        communication_manager.inform(RawEventTypes.NEW_SESSION.name, new_session_event)
+        raise Exception("This should have raised an exception")
+    except Exception as e:
+        assert str(e) == "Communication Manager needs to be configured first!"
+
+def test_communication_manager_can_inform_the_optimizer():
```
georgymh: Please rename to `test_communication_manager_can_inform_new_job_to_the_optimizer` or something more precise.

tests/test_communication_manager.py
neeleshdodda44: > one line

Then can you label exactly where in the code it happens? This can simply be adding numbers to the documentation in the code so that the exact order of operations is more visible.

core/communication_manager.py
```diff
@@ -0,0 +1,156 @@
...
+        # These are unmapped and will only be received when an optimizer is not
+        # active.
+        # NOTE: This should be updated when Gateway PR is merged to
+        # MessageEventTypes.<>.name
+        self.RAW_EVENT_TYPE_2_CALLBACK = {
+            RawEventTypes.NEW_SESSION.name: self._create_session,
+            RawEventTypes.NEW_WEIGHTS.name: self._do_nothing,
+            RawEventTypes.TERMINATE.name: self._do_nothing,
+            RawEventTypes.NOTHING.name: self._do_nothing,
```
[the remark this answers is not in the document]
kiddyboots216: Gateway tries to parse the TX, it's a bad TX, gives it to the Communication Manager anyways.
(see 'delegating error handling to hosts' philosophy in networking)

core/utils/enums.py
```diff
@@ -0,0 +1,102 @@
...
+
+    """
+    SCHEDULE_JOB = "SCHEDULE_JOB"
+    TERMINATE = "TERMINATE"
+    NOTHING = "NOTHING"
+
+def callback_handler(
+    callback_type,
+    callback_dict,
+    default=RawEventTypes.NOTHING.value
```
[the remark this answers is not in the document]
kiddyboots216: done

core/fed_avg_optimizer.py
```diff
@@ -0,0 +1,157 @@
...
+		return callback(payload)
+	
+	def _handle_job_done(self, result):
+		"""
+		Level 1 Callback to handle the payload of a DMLResult
+		Parses job.job_type into Level 2 Callback
+		Also updates the weights
+		"""
+		assert isinstance(result, DMLResult)
+		# assume success for now
```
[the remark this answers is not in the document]
kiddyboots216: done & done

core/fed_avg_optimizer.py
kiddyboots216: sure? you initially were the one who wanted to have these default "NOTHING" values so idrc, removed

core/utils/dmljob.py
kiddyboots216: done

core/utils/enums.py
kiddyboots216: done & done

core/utils/enums.py
kiddyboots216: idk removed

tests/test_communication_manager.py
kiddyboots216: done

kiddyboots216:
Optimizer is tested in the communication manager tests -separating out the tests is not really worth the effort; it's an **incredibly simple** class. 
Addressed all other requests by @georgymh

georgymh:
@kiddyboots216: I disagree that the optimizer is an "incredibly simple class." I think the Dataset Manager, Scheduler, and Runner are lot less complex than the Optimizer and they all have tests. Also, what you have are integration tests between the Communication Manager and the Optimizer. We need **unit tests**.

Some ideas of what to unit test:

- kickoff returns what's expected
- all different types of asks (individually, IOW, touching all possible level 1 and level 2 combinations) return what's expected

core/utils/enums.py
georgymh: Easy: 

1 - Create a mock Communication Manager and use that for the Scheduler's unit tests. Code:

```
class MockCommunicationManager:
    def inform(self, dummy1, dummy2):
        pass
```

2- (Bonus) Make the Scheduler have a `configure()` method where the Configuration Manager is passed. Let's actually make it a pattern that all connections between modules (dependency injections) should be done by a `configure()` method rather than the constructor. 

Now, in high level terms, what are realistic ways the Communication Manager will receive an event that says "NOTHING happened"?

core/communication_manager.py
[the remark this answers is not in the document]
georgymh: Link?

(And note that if this is the only way this happens this should be added in the Gateway PR instead, with tests included)

cc. @nzoghb

core/communication_manager.py
```diff
@@ -80,17 +81,15 @@ def inform(self, event_type, payload):
         if self.optimizer:
             # We have an active session.
             event_type, payload = self.optimizer.ask(event_type, payload)
-            callback = callback_handler(
+            callback = callback_handler_no_default(
```
georgymh: Why did you make this change? Did you mean to do it in L90?

The current design says that the optimizer could inform the Communication Manager to do nothing if that's its policy.

kiddyboots216: true, neither this nor L90 should be w/o default

georgymh: I'm also OK with removing all NOTHINGs everywhere (and `callback_handler()` calls) and not worrying about it for now since we're only supporting one Optimizer at the moment anyways. This would probably make things less error-prone and simpler.

georgymh: Cool, seems we're on the same page with this then.

georgymh: To be clear, what I'm saying is either:

1- L84 using `callback_handler()` and L90 using `callback_handler_no_default()`; or,
2- L84 and L90 using `callback_handler_no_default()`

(If you decide otherwise, then this is dependent on the other thread we're discussing in `enums.py`'s L54 and we should address this thread after that one is solved.)

georgymh:
Yes: writing the tests for the optimizers, writing a "train but not initialize" test for the communication manager, possibly breaking up `RawEventTypes` further for clarity, and updating docstrings if necessary.

kiddyboots216:
No "train but not initialize" test for communication manager (communication manager should never have any say over that logic; it relies on the optimizer, and the optimizer's kickoff will overwrite the job_type of the job it gets to INIT, then replace the weights in the job that it got -this is to allow the optimizer to receive initialization info from ANY node not just the Notebook, good for p2p down the road etc.)
Added tests for the optimizers
Don't think RawEventTypes needs to be broken up until after Gateway PR is merged and we see what the Enums look like with MessageTypes in

core/communication_manager.py
```diff
@@ -79,7 +79,7 @@ def inform(self, event_type, payload):
         if self.optimizer:
             # We have an active session.
             event_type, payload = self.optimizer.ask(event_type, payload)
-            callback = callback_handler_no_default(
+            callback = callback_handler(
```
georgymh: @kiddyboots216: Next time someone makes a string of commits to a branch you're working on, don't "solve merge conflicts" because there's nothing to solve. There are local changes you need to discard like this one.

Fixing.

georgymh:
Will take a look and decide on those later.

core/utils/dmljob.py
neeleshdodda44: Don't think it's worth unresolving, but I believe both of my suggestions cover that @georgymh (`if serialized_job.get('weights', None)` will return `False` if the weights are not in `serialized_job` or if the weights are `None`.

georgymh:
Alright, I made the changes I was going to make (reflected in the Trello card "Finalize PR #26).

There are some commented code I left because I'm not sure if @kiddyboots216 backed it up and I didn't want to make him redo his work.

@neeleshdodda44, please review and I will make any necessary changes ASAP.

core/utils/dmljob.py
georgymh: Hmm, you're right. Making the change.

core/utils/enums.py
```diff
@@ -5,98 +5,120 @@
...
     """
-    Enum to implement Optimizer -> Communication Manager callback pattern
+    Actionable event types that the service can take action on.
 
-    This is essentially a 'command' for the Comm. Mgr. so these are all imperatives.
+    The different event types are the following:
 
-    SCHEUDLE_JOB: Optimizer wants Communication Manager to tell its
-    Scheduler to schedule a job
+        SCHEUDLE_JOB: Instruction to the Execution Pipeline to schedule a job
```
neeleshdodda44: s/SCHEUDLE_JOB/SCHEDULE_JOB

PR #27, opened 2018-10-21
Check data schema

neeleshdodda44:
Major changes:
- Dataset Manager checks raw data so that each dataset has a valid CSV format and that it has a header (assume column names are string). Sidenote: if data is all strings, how can we detect whether it has a header?
- Tests that ensure validation fails when DSM is given raw data without a header or raw data in an invalid CSV format.
- Cleaned up some code in DatasetManager

### DIFF
--- a/core/dataset_manager.py
+++ b/core/dataset_manager.py
@@ -111,10 +111,12 @@ class DatasetManager():
                     dataset = pd.read_csv(file_path, index_col=False)
                 except Exception as e:
                     logging.error(str(e))
-                    assert False, format_message.format(
-                        file=file,
-                        folder=folder,
-                        message=str(e)
+                    raise Exception(
+                        format_message.format(
+                            file=file,
+                            folder=folder,
+                            message=str(e)
+                        )
                     )
                 is_str = lambda c: not c.replace('.','',1).isdigit()
                 assert all([is_str(c) for c in dataset.columns]), \
--- a/tests/test_dataset_manager.py
+++ b/tests/test_dataset_manager.py
@@ -110,14 +110,15 @@ def test_no_header(bad_config_manager_header):
     """
     error_message = ("No header has been provided in file {file} in folder "
                      "{folder}")
+    error_message = error_message.format(
+                        file='test1.csv',
+                        folder='test1',
+                    )
     try:
         dsm = DatasetManager(bad_config_manager_header)
-        assert False, "Error should have been thrown for lack of header"
+        raise Exception("Error should have been thrown for lack of header")
     except AssertionError as e:
-        assert str(e) == error_message.format(
-                            file='test1.csv',
-                            folder='test1',
-                         )
+        assert str(e) == error_message, "Wrong assertion was thrown!"
 
 def test_bad_format(bad_config_manager_format):
     """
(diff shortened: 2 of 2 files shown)


core/dataset_manager.py
```diff
@@ -82,8 +82,48 @@ def __init__(self, config_manager):
...
+            files = os.listdir(folder_path)
+            for file in files:
+                if not file.endswith(".csv"): continue
+                if file[:2] == 'md': continue
+                file_path = os.path.join(folder_path, file)
+                try:
+                    dataset = pd.read_csv(file_path, index_col=False)
+                except Exception as e:
+                    logging.error(str(e))
+                    assert False, format_message.format(
```
georgymh: Can we raise an exception instead?

tests/test_dataset_manager.py
```diff
@@ -88,6 +104,39 @@ def drop_duplicates(df):
...
 
+def test_no_header(bad_config_manager_header):
+    """
+    Test that DSM invalidates dataset with no header.
+    """
+    error_message = ("No header has been provided in file {file} in folder "
+                     "{folder}")
+    try:
+        dsm = DatasetManager(bad_config_manager_header)
+        assert False, "Error should have been thrown for lack of header"
```
georgymh: Can you raise an exception here too?

tests/test_dataset_manager.py
```diff
@@ -88,6 +104,39 @@ def drop_duplicates(df):
...
+    except AssertionError as e:
+        assert str(e) == error_message.format(
```
georgymh: Should this assert raise a meaningful error if the condition fails?

tests/test_dataset_manager.py
```diff
@@ -88,6 +104,39 @@ def drop_duplicates(df):
...
+    """
+    Test that DSM invalidates dataset with invalid CSV format.
+    """
+    format_message = ("The file {file} in folder {folder} was improperly "
+                      "formatted. Please refer to the following error "
+                      "message from pandas for more information: {message}")
+                      
+    try:
+        dsm = DatasetManager(bad_config_manager_format)
+        assert False, "Error should have been thrown for bad format"
```
georgymh: Same as above.

tests/test_dataset_manager.py
```diff
@@ -88,6 +104,39 @@ def drop_duplicates(df):
...
+    except AssertionError as e:
+        assert str(e) == format_message.format(
```
georgymh: Same as above.

core/dataset_manager.py
neeleshdodda44: Sure.

tests/test_dataset_manager.py
neeleshdodda44: Sure.

tests/test_dataset_manager.py
neeleshdodda44: Yeah, will add.

tests/test_dataset_manager.py
neeleshdodda44: Sure.

tests/test_dataset_manager.py
neeleshdodda44: Sure.

neeleshdodda44:
Pretty much all of the algorithms I've seen that detect headers basically check if the first line has comma delimited strings. So tl;dr, I'm not aware of a fix as of now, but we can worry about this later.

Maybe this could be designated as an issue for the future?

PR #29, opened 2018-10-27
Averaging

kiddyboots216:
Adding averaging jobtype to Runner.
Also added a mock Communication Jobtype just to get to the Averaging test.
Please do me an approve friendo.
No am many lines code change.
(P.S. the integration test for the Communication Manager looks horrific, I know, but just imagine how hard it was to make the test!)


georgymh:
Ok, since it's not dependent on the Blockchain code, let's remove the code related to it (NEW_INFO, the code I removed from the optimizers in PR #26, etc).

About the COMM code, it sounds like there's less boilerplate/temp code to be written in this PR if the COMM PR is submitted first, so why don't we get the COMM PR out first and then we come back to this one? It will make reviewing both/ approving both easier for both of us.

kiddyboots216:
Ok, this PR will have to wait on the Gateway PR and then the COMM PR then.
NEW_INFO, etc. is not dependent on the Blockchain code. So it should stay in regardless.

kiddyboots216:
This PR is ready to be merged. @neeleshdodda44 Please review.

core/fed_avg_optimizer.py
```diff
@@ -52,16 +52,23 @@ def __init__(self, initialization_payload):
 		serialized_job = initialization_payload.get('serialized_job')
 		self.job = deserialize_job(serialized_job)
 		optimizer_params = initialization_payload.get('optimizer_params')
-		self.listen_iterations = optimizer_params.get('listen_iterations')
-		self.listen_bound = optimizer_params.get('listen_bound')
+		self.listen_iterations = optimizer_params.get('listen_iterations', 0)
+		self.listen_bound = optimizer_params.get('listen_bound', 2)
```
neeleshdodda44: I would consider moving the default value of 2 of `listen_bound` to the `configuration.ini`, since it seems arbitrary.

tests/test_communication_manager.py
```diff
@@ -72,7 +96,33 @@ def test_communication_manager_can_initialize_and_train_model():
         scheduler.runners_run_next_jobs()
         time.sleep(0.1)
     assert len(scheduler.processed) == 2
-
+    while len(scheduler.processed) == 2:
```
neeleshdodda44: This should timeout after a short amount of time, so that tests don't hang in case stuff breaks.

tests/test_communication_manager.py
```diff
@@ -72,7 +96,33 @@ def test_communication_manager_can_initialize_and_train_model():
...
+        "key": MessageEventTypes.NEW_WEIGHTS.name,
+        "content": {
+            "weights": serialize_weights(communication_manager.optimizer.job.weights)
+        }
+    }
+    communication_manager.inform(
+        RawEventTypes.NEW_INFO.name,
+        new_weights_event
+    )
+    while len(scheduler.processed) == 3:
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
```diff
@@ -72,7 +96,33 @@ def test_communication_manager_can_initialize_and_train_model():
...
+        scheduler.runners_run_next_jobs()
+        time.sleep(0.1)
+    assert communication_manager.optimizer.listen_iterations == 1
+    communication_manager.inform(
+        RawEventTypes.NEW_INFO.name,
+        new_weights_event
+    )
+    while len(scheduler.processed) == 4:
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
```diff
@@ -110,19 +121,31 @@ def test_communication_manager_can_initialize_and_train_and_average_model():
...
         new_weights_event
     )
-    while len(scheduler.processed) == 3:
+    timeout = time.time() + 3
+    while time.time() < timeout and len(scheduler.processed) == 3:
+        # averaging job
         scheduler.runners_run_next_jobs()
         time.sleep(0.1)
+    # we've only heard one set of new weights so our listen_iters are 1
     assert communication_manager.optimizer.listen_iterations == 1
```
neeleshdodda44: Same as above, can't tell whether assert statement fails because of timeout or runner error.

tests/test_communication_manager.py
```diff
@@ -110,19 +121,31 @@ def test_communication_manager_can_initialize_and_train_and_average_model():
...
         new_weights_event
     )
-    while len(scheduler.processed) == 4:
+    timeout = time.time() + 3
+    while time.time() < timeout and len(scheduler.processed) == 4:
+        # second averaging job
         scheduler.runners_run_next_jobs()
         time.sleep(0.1)
+    # we've heard both sets of new weights so our listen_iters are 2
     assert communication_manager.optimizer.listen_iterations == 0
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
```diff
@@ -89,17 +89,28 @@ def test_communication_manager_can_initialize_and_train_and_average_model():
...
+    timeout = time.time() + 3
+    while time.time() < timeout and len(scheduler.processed) == 2:
+        # communication job
         scheduler.runners_run_next_jobs()
         time.sleep(0.1)
     assert len(scheduler.processed) == 3
+    # now the communication manager should be idle
+    scheduler.runners_run_next_jobs()
+    time.sleep(0.1)
+    assert len(scheduler.processed) == 3
```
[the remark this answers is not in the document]
neeleshdodda44: EDIT I see some assertions on length after some of the jobs, but you should still address having messages and explicit assertions for timeouts and asserting the length after the initialization job.

Also nitpick but tests could be stronger if you actually verify that the processed job matches what you expect it to be (i.e. tests would pass if three initialization jobs were run?)

core/runner.py
```diff
@@ -3,18 +3,18 @@
...
 from data.iterators import create_random_train_dataset_iterator
 from data.iterators import create_random_test_dataset_iterator
 from core.utils.keras import train_keras_model, validate_keras_model
-from core.utils.keras import serialize_weights
+from core.utils.keras import serialize_weights, deserialize_weights
 from core.utils.dmlresult import DMLResult
 from core.utils.enums import JobTypes, callback_handler_no_default
-# from core.utils.federated_learning_utils import federated_averaging
-#       NOTE: Commented until the next PR is done.
+
```
neeleshdodda44: Nitpick: delete this extra line.

core/runner.py
```diff
@@ -221,3 +225,43 @@ def _initialize_model(self, job):
...
+            status='successful',
+            job=job,
+            results={
+                'weights': averaged_weights,
+            },
+            error_message="",
+        )
+        return result
+    
+    def weighted_running_avg(self, sigma_x_i_div_w_i, x_n, sigma_w_i, w_n):
```
neeleshdodda44: Nitpick but documentation should be more clear on what the arguments are (i.e. explicitly state that the arguments are numpy arrays, what the other two arguments represent, etc.)

tests/test_dmlrunner.py
```diff
@@ -69,3 +69,22 @@ def test_dmlrunner_validate_job_returns_stats(config_manager):
...
     assert type(val_stats) == dict
+
+def test_dmlrunner_initialize_job_weights_can_be_serialized(config_manager):
+    model_json = make_model_json()
+    runner = DMLRunner(config_manager)
+    initialize_job = make_initialize_job(model_json)
+    initial_weights = runner.run_job(initialize_job).results['weights']
+    same_weights = deserialize_weights(serialize_weights(initial_weights))
+    assert(np.shape(same_weights) == np.shape(initial_weights))
+    assert all([np.count_nonzero(arr)==0 for arr in np.subtract(same_weights, initial_weights)])
```
neeleshdodda44: We might have talked about this offline, but why not directly check if both arrays are equal? Try [this](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.array_equal.html), and I believe this reduces the complexity of these two assert statements.

core/runner.py
kiddyboots216: Check latest commit

core/runner.py

tests/test_communication_manager.py
kiddyboots216: Done

tests/test_communication_manager.py
kiddyboots216: Done

tests/test_communication_manager.py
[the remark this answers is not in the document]
kiddyboots216: See latest commit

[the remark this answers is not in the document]
kiddyboots216: Re: Runner errors, runner errors should be caught and logged so that should help the developer with debugging.

tests/test_dmlrunner.py
```diff
@@ -69,3 +69,22 @@ def test_dmlrunner_validate_job_returns_stats(config_manager):
...
+
+def test_dmlrunner_averaging_weights(config_manager):
+    model_json = make_model_json()
+    runner = DMLRunner(config_manager)
+    initialize_job = make_initialize_job(model_json)
+    initial_weights = runner.run_job(initialize_job).results['weights']
+    serialized_weights = serialize_weights(initial_weights)
+    initialize_job.set_weights(initial_weights, serialized_weights, 1, 1)
+    averaged_weights = runner._average(initialize_job).results['weights']
+    assert all([np.count_nonzero(arr)==0 for arr in np.subtract(averaged_weights, initial_weights)])
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
```diff
@@ -89,27 +89,34 @@ def test_communication_manager_can_initialize_and_train_and_average_model():
...
+    assert len(scheduler.processed) == 2, "Training not completed in time!"
+    assert communication_manager.optimizer.job.job_type == JobTypes.JOB_COMM.name, \
+        "Should be ready to communicate!"
     timeout = time.time() + 3
     while time.time() < timeout and len(scheduler.processed) == 2:
         # communication job
         scheduler.runners_run_next_jobs()
         time.sleep(0.1)
-    assert len(scheduler.processed) == 3
+    assert len(scheduler.processed) == 3, "Communcation not completed in time!"
```
neeleshdodda44: Nitpick: s/Communcation/Communication

tests/test_dmlrunner.py
neeleshdodda44: Sorry, try [this](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.allclose.html) . 

```
assert all(np.allclose(arr1, arr2) for arr1,arr2 in zip(same_weights, initial_weights)) 
```

neeleshdodda44: Makes it more clear what the assert statements are trying to accomplish.

kiddyboots216: Done

tests/test_dmlrunner.py
kiddyboots216: Done

tests/test_communication_manager.py
kiddyboots216: Done

tests/test_communication_manager.py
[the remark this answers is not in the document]
neeleshdodda44: You're right, but if the issue was originally Runner errors, then the outcome would still be a timeout (since failed jobs are discarded after three tries). 

So then I think you should clarify in the messages that the issue could be an actual timeout or Runner error (since it's currently impossible to determine which is which?)

tests/test_communication_manager.py
```diff
@@ -0,0 +1,282 @@
...
+        new_session_event
+    )
+    assert communication_manager.optimizer.job.job_type == JobTypes.JOB_INIT.name, \
+        "Should be ready to init!"
+    timeout = time.time() + 3
+    while time.time() < timeout and len(scheduler.processed) == 0:
+        # initialization job
+        scheduler.runners_run_next_jobs()
+        time.sleep(0.1)
+    assert len(scheduler.processed) == 1, "Initialization not completed in time!"
```
neeleshdodda44: _Carrying over previous thread since it was outdated, will resolve that thread_

You're right, but if the issue was originally Runner errors, then the outcome would still be a timeout (since failed jobs are discarded after three tries).

So then I think you should clarify in the messages that the issue could be an actual timeout or Runner error (since it's currently impossible to determine which is which?)

tests/test_communication_manager.py
```diff
@@ -0,0 +1,282 @@
...
+    assert communication_manager.optimizer.job.job_type == JobTypes.JOB_TRAIN.name, \
+        "Should be ready to train!"
+    timeout = time.time() + 3
+    while time.time() < timeout and len(scheduler.processed) == 1:
+        # training job
+        scheduler.runners_run_next_jobs()
+        time.sleep(0.1)
+    assert len(scheduler.processed) == 2, "Training not completed in time!"
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
```diff
@@ -0,0 +1,282 @@
...
+    assert communication_manager.optimizer.job.job_type == JobTypes.JOB_COMM.name, \
+        "Should be ready to communicate!"
+    timeout = time.time() + 3
+    while time.time() < timeout and len(scheduler.processed) == 2:
+        # communication job
+        scheduler.runners_run_next_jobs()
+        time.sleep(0.1)
+    assert len(scheduler.processed) == 3, "Communication not completed in time!"
```
neeleshdodda44: Same as above.

tests/test_communication_manager.py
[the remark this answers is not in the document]
kiddyboots216: You can determine whether it's a timeout or Runner error by inspecting the log; if there are Runner errors and it keeps retrying, then it will show up in the log. If the test times out and there are no Runner errors, then it's a timeout. I've added this to the test docstring.

core/runner.py
neeleshdodda44: If this is done, how come I still see the extra line -_- (not trying to be sassy)

tests/test_communication_manager.py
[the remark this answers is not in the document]
neeleshdodda44: I meant programmatically, so that assertion would be able to differentiate the actual error in its message.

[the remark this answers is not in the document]
neeleshdodda44: Resolving, we can discuss this further (if necessary) in the other thread

PR #30, opened 2018-11-05
Transform+Split in Runner

neeleshdodda44:
Major changes: 

**Runner**
- Transform and training-test split now takes places as a new type of job: `transform_and_split` in the method `_transform_and_split_data`. 
- Corresponding enums have been created
- Tests have been updated to test transform/split functionality, as well as test training/testing off of transformed and split data.

**DMLJob**
- New attributes added for `raw_filepath` (previously `rfp`), `session_filepath`, and `transform_function`. 
- `serialize_job` and `deserialize_job` have been updated accordingly.

**Testing utils**
- Created a new utils for transforming and splitting data.
- For tests that facilitate training + validation outside of Runner tests, I just had added the session filepath manually to the initialize job. Although transform and splitting for all of these tests is not necessary, **we need to implement transforming and splitting data in the optimizer logic**. This should be a separate PR, maybe assigned to @kiddyboots216?

**Dataset Manager**
- Refactored code, tests and documentation so that it is clear the Dataset Manager does not handle transforms any more.


core/runner.py
```diff
@@ -166,12 +176,23 @@ def _validate(self, job, custom_iterator=None):
...
+            "Transformed training and test sets not created yet!"
+        train_dataset_path = os.path.join(
+                                job.session_filepath, 
+                                'train.csv'
+                            )
+        test_dataset_path = os.path.join(
+                                job.session_filepath, 
+                                'test.csv'
+                            )
+        data_count = count_datapoints(job.session_filepath)
```
kiddyboots216: since it's here twice can we make this a helper function? def _helper_dataset() returns train_dataset_path, test_dataset_path, data_count. For lines 181-189 in this function, and these also appear in _train.

core/runner.py
```diff
@@ -221,3 +242,99 @@ def _initialize_model(self, job):
...
+        # 7. Return job with assigned session folder filepath
+        results = DMLResult(
+                    status='successful',
+                    job=job,
+                    results = {
+                        'session_filepath': session_filepath
+                    },
+                    error_message=""
+                )
+        return results
```
kiddyboots216: How are you ensuring that this session_filepath is actually what will be used throughout the rest of the session? Are you just saying that the session_filepath will remain in the DMLJob, which is persisting inside of the FedAvgOptimizer for the duration of the session?

tests/test_dmlrunner.py
```diff
@@ -19,6 +23,42 @@ def config_manager():
...
+def small_filepath():
+    return 'tests/artifacts/runner_scheduler/test'
+
+def accurate_transform_test(dsm, expected_test_transformed, split):
+    """
+    Check that the data in this folder is the result of calling the transform function on
+    each csv in the raw data filepath
+    """
+    transform_dsm = dsm.get_transformed_data()
+    keys = list(transform_dsm.keys())
```
kiddyboots216: doesn't look like this variable is used anywhere?

tests/test_dmlrunner.py
```diff
@@ -19,6 +23,42 @@ def config_manager():
...
+    train = transform_dsm['train']
+    test = transform_dsm['test']
+    count = len(expected_test_transformed)
+    assert len(train) == int(split * count)
+    assert len(test) == int((1 - split) * count)
+    actual_test_transformed = train.append(test)
```
kiddyboots216: Isn't this just -the original data now?

tests/test_dmlrunner.py
```diff
@@ -19,6 +23,42 @@ def config_manager():
...
+    actual_test_transformed = actual_test_transformed.reset_index(drop=True)
+    merged_data = pd.merge(
+        actual_test_transformed,
+        expected_test_transformed,
+        on = ['a', 'b', 'c'],
+        how = 'inner'
+    )
+    assert len(actual_test_transformed) == count
+    assert len(merged_data) == count
```
kiddyboots216: It seems like it is easier for this test to check element-wise, order-agnostic equality.

core/runner.py
neeleshdodda44: Actually, the `session_filepath` is just returned as a part of `results`. I would assume that in the optimizer somewhere, there's a method like: 

```
def _done_transform_split(self, dmlresult_obj):
...
self.job.session_filepath = dmlresult_obj.results['session_filepath']
...
```

tests/testing_utils.py
```diff
@@ -26,37 +26,55 @@ def make_initialize_job(model_json):
...
+        model_json,
+        "keras",
+        hyperparams=make_hyperparams(),
+        label_column_name='label',
+        raw_filepath=raw_filepath,
+        transform_function=transform_function
+    )
+    return transform_split_job
+
+def make_hyperparams(split=0.004):
```
kiddyboots216: why is split a 'hyperparameter'? Hyperparameters are things that are adjusted in hyperparameter search/sweep, i.e. batch size, learning rate, epochs

core/runner.py
neeleshdodda44: Anticipating your followup: (@georgymh can correct me if I'm wrong) but jobs should only be altered outside of the scheduler, since we encountered issues with multithreading and shared memory when we altered the job inside of the Runner.

tests/test_dmlrunner.py
neeleshdodda44: Not using this function, will delete.

tests/test_dmlrunner.py
neeleshdodda44: Not using this function, will delete.

tests/test_dmlrunner.py
neeleshdodda44: Not using this function, will delete.

core/runner.py
kiddyboots216: That's fine, I just wanted to make sure that this was your intent because I'll need to modify the optimizer. Matter of fact, if the optimizer isn't doing this currently, how are the tests passing? Since the optimizer isn't calling for a transform_and_split job, and there are assert statements in the train job.

tests/testing_utils.py
neeleshdodda44: Not my design decision, you should ask @georgymh.

core/runner.py
neeleshdodda44: See PR description

core/runner.py
neeleshdodda44: Done.

neeleshdodda44:
@georgymh Can you review this PR soon so I don't have to keep fixing merge conflicts from recently merged PRs 😢

core/runner.py
```diff
@@ -81,6 +80,21 @@ def run_job(self, job):
         logging.info("Finished running job! (type: {0})".format(job.job_type))
         return results
 
+    def _set_up(self, job):
```
georgymh: Doing `count_datapoints()` on every job seems expensive. Could we do this once during `_transform_and_split_data()` and return it in the results?

core/utils/dmljob.py
```diff
@@ -18,7 +19,10 @@ def __init__(
...
         weights=None,
         hyperparams=None,
-        label_column_name=None
+        label_column_name=None,
+        raw_filepath=None,
+        session_filepath=None,
+        transform_function=None
         ):
         """
         Initializes a DML Job object.
```
georgymh: Can we add docs for the new properties?

tests/test_dmlrunner.py
```diff
@@ -2,14 +2,18 @@
```
georgymh: We should add a test that tests a split where the user doesn't want to split and wants to train on all the data.

PR #31, opened 2018-11-10
Communication job

kiddyboots216:
This pull requests fleshes out the communication job functionality.

Adds a small bevvy of changes to blockchain_client, communication_manager, and fed_avg_optimizer to make these changes possible.

This PR is NOT READY FOR REVIEW. I am making this PR to use Github's interface for resolving merge conflicts.


kiddyboots216:
"I would consider keeping the .travis.yml comments. They're useful."
You...literally asked them to be removed in the Google doc...but sure.

"Since this PR is adding the Blockchain Gateway to the bootstrapper, can we have instructions on how to run the dml command locally and tinker with the BG? (Ideally we're able to replicate the demo you did during the last meeting.)"
The instructions are already in the README on Master.

"Look into removing set_weights and set_keys in the DMLJob."
Definitely should do this.

"Explain the rationale of adding do_nothing(), NEW_SESSION, and TERMINATE to the Optimizer."
Rationale is already explained in the docstring.

"Can we do some error checking in the runner's _communicate() routine? What if the setter doesn't return a valid tx_receipt? Should that instead return a failed DMLResult?"
tx_receipt contains errors in consensus, as for actual errors in the service that needs to be handled with the general changes to Runner error-handling. So I don't think so.

"Make L63 of the integration tests more explicit."
Can you elaborate?

Everything else addressed with the exception of scheduling communication_job, which as it turns out doesn't work very well. I've tried debugging this for a while but can't make any headway.

georgymh:
>The instructions are already in the README on Master.

It's incomplete though. You should add instructions/snippets to submit DML requests through the Blockchain Gateway after running `dml`.



> Rationale is already explained in the docstring.

I checked again and didn't see any rationale. I will just answer it here for documentation purposes (and so that other reviewers are aware of the whys):

- NEW_SESSION is added because the Blockchain Gateway may pick up a new session message in the middle of a session and we want the service to just ignore it. Hence the `_do_nothing()`.

- TERMINATE was just an addition we were already planning to do. The Blockchain Gateway may pick up a termination message, in which case the Optimizer will instruct the Communication Manager to end the session (and destroy the Optimizer).



>"Make L63 of the integration tests more explicit."
Can you elaborate?

Use something like:

```python
function(arg1=a, arg2=b, arg3=c)
```

But properly indented.

georgymh:
We should be able to mock that, no? Is there a way to have only one node in the network and do the same HTTP request that explora would be doing to set something on the blockchain?

I think on your demo you were starting with a blockchain that already had a job/DML request in it. That would work too. The point is to allow us to do more than just run `dml` and patiently wait for nothing.

kiddyboots216:
The easiest way is to create a pytest which sets exactly what you want to the chain, run that pytest **only** and then start the DML service without resetting the chain.

georgymh:
Rather than a test, could we have a utils we could execute from command line? I'm also ok with calling `curl` manually. We just need instructions on the README.

kiddyboots216:
No, it's not feasible. You would have to make the model_json, make a DMLJob, serialize the job, make a session, put the session into ipfs, get the content hash, and then put that onto the blockchain. Calling curl is doable, sure, but the blockchain is going to change anyway, so this is not a good use of time for this PR.

georgymh:
Ok, sounds like you ended up updating the README anyways.

What's `<your DML session here>`? Can you put an example DML session there? Or make the function calls necessary to create one?

georgymh:
Cool. Thanks for "wasting" your time in 2578551 as a way to avoid the team actually ***wasting*** its collective time trying to make code they didn't write work. In fact, let's "waste" our time more often.

On a more serious note though, a few of the points I put above weren't addressed (like the travis file one, the .gitignore one, removing the DMLJob functions/explaining why they should be there, and adding docstrings for the new properties in the DMLJob constructor). Please address them and I will take a look at the tests tomorrow. They look comprehensive, which is good.

core/bootstrapper.py
```diff
@@ -4,7 +4,7 @@
 from core.dataset_manager import DatasetManager
 from core.communication_manager import CommunicationManager
 from core.scheduler import DMLScheduler
-
+from core.blockchain.blockchain_gateway import BlockchainGateway
```
neeleshdodda44: Please add the extra line back in.

core/communication_manager.py
```diff
@@ -1,10 +1,10 @@
...
-from core.utils.enums       import RawEventTypes, ActionableEventTypes
-from core.utils.enums       import callback_handler_no_default
-from core.fed_avg_optimizer import FederatedAveragingOptimizer
-from core.utils.dmljob      import DMLJob
-
+from core.utils.enums                   import RawEventTypes, MessageEventTypes, ActionableEventTypes
+from core.utils.enums                   import callback_handler_no_default
+from core.fed_avg_optimizer             import FederatedAveragingOptimizer
+from core.utils.dmljob                  import DMLJob
+from core.blockchain.blockchain_utils   import TxEnum
```
neeleshdodda44: Same as above.

tests/test_fed_avg_optimizer.py
```diff
@@ -109,3 +118,10 @@ def test_optimizer_schedules_communication_after_training(initialization_payload
     assert event_type == ActionableEventTypes.SCHEDULE_JOBS.name
     for job in job_arr:
         assert job.job_type == JobTypes.JOB_COMM.name
+
+def test_cleanup():
```
neeleshdodda44: Although convenient since it runs after all of the tests, I wouldn't make clean up a test since it's just something that needs to be done rather than a functionality of the Optimizer you're actually testing.

Tl;dr s/test_cleanup/cleanup, remove the assert statement, and call this method at the end of the relevant tests. If this seems repetitive, you might want to consider using something similar to teardown functionality, see [here](https://stackoverflow.com/questions/22627659/run-code-before-and-after-each-test-in-py-test)

kiddyboots216: I had the exact functionality you're describing before. It didn't work since tests that failed wouldn't get to their cleanup code, which means that developers have to rm the transformed data every time. I will look into using the fixture.

core/bootstrapper.py
kiddyboots216: Done in upcoming commit

core/communication_manager.py
kiddyboots216: Done in upcoming commit

core/fed_avg_optimizer.py
```diff
@@ -147,57 +154,73 @@ def _done_training(self, dmlresult_obj):
 		weights.
 		"""
 		new_weights = dmlresult_obj.results.get('weights')
+		self.job.omega = dmlresult_obj.results.get('omega')
+		self.job.sigma_omega = self.job.omega
 		self._update_weights(new_weights)
 		self.job.job_type = JobTypes.JOB_COMM.name
+		self.job.key = "test"
```
neeleshdodda44: Was this intentional? If so, I assume the key won't always be `test`, so how would this normally be set?

kiddyboots216: Key management is coming in next blockchain PR, as has been discussed on #dev.

core/questions.csv
```diff
@@ -4,6 +4,10 @@ How many runners should the scheduler manage?,SCHEDULER,num_runners,4
 How often should the scheduler run in minutes?,SCHEDULER,frequency_in_mins,1
 Where is the weights directory?,RUNNER,weights_directory,weights
 What is the max number of times a job should run?,SCHEDULER,max_tries,3
+What is the port for IPFS?,BLOCKCHAIN,ipfs_port,5001
+What is the port for HTTP?,BLOCKCHAIN,http_port,3000
+What is the host for HTTP?,BLOCKCHAIN,host,127.0.0.1
+What is the timeout for HTTP?,BLOCKCHAIN,timeout,5
```
neeleshdodda44: Are these questions that data provider will actually answer? If not, please change the question entry in all of them to `SECRET`.

tests/test_fed_avg_optimizer.py
kiddyboots216: Fixture doesn't work. I'm going to leave as is until an actual solution can be found.

core/fed_avg_optimizer.py
neeleshdodda44: Can you at least add a TODO so it's clear this will be changed?

core/questions.csv
kiddyboots216: The first three will be answered, I think. Last changed to SECRET. Easy enough to change in future.

core/runner.py
```diff
@@ -320,8 +349,12 @@ def _average(self, job):
         """
         Average the weights in the job weighted by their omegas.
         """
-        deserialized_new_weights = deserialize_weights(job.new_weights)
-        averaged_weights = self._weighted_running_avg(job.weights, deserialized_new_weights, job.sigma_omega, job.omega)
+        # TODO: Figure out how to assert that this job actually has weights/new_weights.
+        # assert isinstance(job.weights, np.ndarray), "No weights supplied to average!"
+        # assert isinstance(job.new_weights, np.ndarray), "No new_weights supplied to average!"
```
neeleshdodda44: Could you briefly explain the TODO?

kiddyboots216: Basically I want
`assert job.new_weights`
But that gives me 'truth value of array is ambiguous'

core/utils/dmljob.py
```diff
@@ -47,31 +47,28 @@ def __init__(
...
         """
         self.job_type = job_type
         self.serialized_model = serialized_model
         self.framework_type = framework_type
         self.weights = weights
         self.hyperparams = hyperparams
         self.label_column_name = label_column_name
+        self.omega = None
+        self.sigma_omega = None
+        self.new_weights = None
```
neeleshdodda44: Any reason why you chose to always set these variables to `None` instead of having them set to `None` as a default option (like other attributes)?

kiddyboots216: Yes. Omega, sigma_omega, and new_weights can only be set after the initialization of a job.

core/questions.csv
neeleshdodda44: Sounds good. Not 100% sure whether Config Manager supports sections whose questions are only partially secret, so just a heads up (trivially you could just change the name of the section for the last question to `BLOCKCHAIN_SECRET` if it doesn't work.

core/runner.py
neeleshdodda44: You're basically checking that it's not empty right? Why not convert to `list` and then assert?

tests/test_all_integration.py
```diff
@@ -0,0 +1,335 @@
...
+def mnist_filepath():
+    return 'tests/artifacts/integration/mnist'
+
+@pytest.fixture
+def new_session_event(mnist_filepath):
+    # initialize_job = make_initialize_job(make_model_json())
+    # initialize_job.hyperparams['epochs'] = 10
+    # initialize_job.hyperparams['batch_size'] = 128
+    # initialize_job.hyperparams['split'] = .9
+    # serialized_job = serialize_job(initialize_job)
```
neeleshdodda44: If you're not going to use this anymore, you can go ahead and delete this.

kiddyboots216: Done in upcoming commit

tests/test_fed_avg_optimizer.py
neeleshdodda44: That's a good point (I've had the same issue), but I still think that it shouldn't be a test when you're merging. Perhaps you can do clean up at end of methods and make this function available in testing utils. That way, when people start developing they can just copy over the general function, modify for their usage, and then delete for PR.

If you still disagree, then let's make this a card and figure it out later.

tests/test_dmlrunner.py
```diff
@@ -31,16 +33,70 @@ def mnist_filepath():
...
+    # Clean up
+    if session_filepath and os.path.isdir(session_filepath):
+        session_filepaths.add(session_filepath)
+    return result
+
+def test_dmlrunner_communicate_job(config_manager, train_dmlresult_obj):
+    runner = DMLRunner(config_manager)
+    comm_job = train_dmlresult_obj.job
+    comm_job.key = "test"
+    result = runner._communicate(comm_job)
```
neeleshdodda44: Why not `runner.run_job(comm_job)`?

kiddyboots216: Done in upcoming commit

neeleshdodda44:
Good work. Left some comments, I'll review again after you address those.

Please try to keep PRs smaller in the future! (195 commits before I started reviewing :( )

tests/test_configuration_manager.py
```diff
@@ -20,10 +20,11 @@ def comments_test(config_filepath):
     """
     Check that comments were added to secret sections
     """
-    secret_sections = ['DB_CLIENT']
+    secret_sections = ['DB_CLIENT', 'BLOCKCHAIN']
     secret_section_message = "; DO NOT MODIFY THIS SECTION"
     with open(config_filepath, "r") as f:
         lines = [line.strip() for line in f.readlines()]
+        print(lines)
```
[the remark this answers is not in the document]
kiddyboots216: Done

tests/test_fed_avg_optimizer.py
kiddyboots216: Did your thing

core/runner.py
```diff
@@ -39,18 +42,23 @@ class DMLRunner(object):
...
+        Sets up IPFS client for _communicate.
         """
         config = config_manager.get_config()
         self.iden = str(uuid.uuid4())[:8]
-        
         self.config = dict(config.items("RUNNER"))
+        self.client = ipfsapi.connect(
+            config.get('BLOCKCHAIN', 'host'), 
+            config.getint('BLOCKCHAIN', 'ipfs_port'))
+        self.port = config.getint('BLOCKCHAIN', 'http_port')
```
neeleshdodda44: In #33 you set up an IPFS Client in DatasetManager, and now you're setting one up here (they seem to be identical). Not sure if I missed this convo but was this intentional? If not, I think it would be better to set up the client in the Bootstrapper and pass it to the Dataset Manager and the Communication Manager -> Optimizer on instantiation.

kiddyboots216: Yeah this was intentional. But I can do the other thing as well.

tests/test_all_integration.py
```diff
@@ -0,0 +1,316 @@
...
+    config_filepath='tests/artifacts/integration/configuration.ini'
+)
+
+@pytest.fixture
+def mnist_filepath():
+    return 'tests/artifacts/integration/mnist'
+
+@pytest.fixture(scope='session')
+def transformed_filepath():
+    return 'tests/artifacts/integration/mnist'+'/transformed'
```
neeleshdodda44: Why not just put the two strings together?

kiddyboots216: I wanted the first string to not be hardcoded and to be mnist_filepath. But pytest kept yelling at me.

tests/test_dmlrunner.py
```diff
@@ -15,15 +15,24 @@
...
+
+@pytest.fixture(scope='session', autouse=True)
+def cleanup(transformed_filepath, transformed_filepath_2):
+    # Will be executed before the first test
+    yield
+    # Will be executed after the last test
+    if os.path.isdir(transformed_filepath):
+        shutil.rmtree(transformed_filepath)
+    if os.path.isdir(transformed_filepath_2):
+        shutil.rmtree(transformed_filepath_2)
```
neeleshdodda44: litty 

You can go ahead move the Trello Card to complete ig (or just delete since you resolved it in this PR)

kiddyboots216: Moved to complete

tests/test_all_integration.py
neeleshdodda44: More details? Also, why is the scope `session`?

tests/test_all_integration.py
```diff
@@ -0,0 +1,316 @@
...
+        blockchain_gateway_2._filter_new_session_info)
+    # (8) Scheduler_1 and Scheduler_2 runs the following jobs:
+        # (6a) JOB_AVG
+        # (6b) JOB_AVG
+        # (6a) JOB_TRAIN
+        # (6a) JOB_COMM
+    timeout = time.time() + 20
+    while time.time() < timeout and (len(scheduler.processed) != 8\
+        or len(scheduler_2.processed) != 8):
+        time.sleep(4)
```
neeleshdodda44: Curious, why 4 seconds here and 5 seconds in line 134? If these are arbitrary, why not just set them to like 1 second so the tests run faster (unless this causes jobs to run slower or smt)?

tests/test_all_integration.py
kiddyboots216: Scope is session so that it can be used by the cleanup fixture which has scope=session, and scope is session for the cleanup fixture so that it will run after all the other tests. Fixtures with different scopes can't be used together apparently. So that's why it yells at me for using mnist_filepath.

tests/test_all_integration.py
```diff
@@ -0,0 +1,316 @@
...
+    blockchain_gateway_2._listen(blockchain_gateway_2._handle_new_session_info,
+        blockchain_gateway_2._filter_new_session_info)
+    # (13) Optimizer tells Communication Manager to schedule JOB_AVG
+    # (14) Scheduler_1 and Scheduler_2 runs the following jobs:
+        # (9a) JOB_AVG
+        # (9a) JOB_AVG
+    timeout = time.time() + 10
+    while time.time() < timeout and (len(scheduler.processed) != 10\
+        or len(scheduler_2.processed) != 10):
+        time.sleep(2)
```
neeleshdodda44: Same as above.

tests/test_all_integration.py
kiddyboots216: If the thread goes to sleep, it can be interrupted, therefore used by another process. So when it's 1 second jobs run slower.

neeleshdodda44: Makes sense.

tests/test_all_integration.py
kiddyboots216: Same as above

core/runner.py
kiddyboots216: Oh wait yeah the reason I did this instead of Bootstrapper is because if I do it with Bootstrapper only then the Runner/Dataset Manager wouldn't pass any tests unless they were bootstrapped. So I'm actually going to leave it like this for time being.

tests/test_all_integration.py
neeleshdodda44: I see. What about making all of the fixtures session scope? IF not, can you put the two strings together?

tests/test_all_integration.py
```diff
@@ -0,0 +1,316 @@
...
+    scheduler.stop_cron()
+    scheduler_2.stop_cron()
+    assert len(scheduler.processed) == 10, "Jobs failed/not completed in time!"
+    assert len(scheduler_2.processed) == 10, "Jobs failed/not completed in time!"
+    # (10) Optimizer terminates
+    assert communication_manager.optimizer is None, "Should have terminated!"
```
neeleshdodda44: You can just do `not communication_manager.optimizer1

tests/test_all_integration.py
```diff
@@ -0,0 +1,316 @@
...
+    assert communication_manager_2.optimizer is None, "Should have terminated!"
```
neeleshdodda44: Same as above.

tests/test_all_integration.py
```diff
@@ -0,0 +1,316 @@
...
+    # and that completes one local round of federated learning!
+
+def test_communication_manager_can_initialize_and_train_model(mnist_filepath, new_session_event):
```
neeleshdodda44: I see you copied this over, can you just quickly confirm whether anything changed in this test (so if I need to review it or not)?

tests/test_all_integration.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    Integration test that checks that the Communication Manager can initialize,
+    train, (and soon communicate) a model, and average a model.
+
+    NOTE: This should be renamed after the COMM PR.
```
neeleshdodda44: This is the COMM PR?

tests/test_all_integration.py
kiddyboots216: Nope

tests/test_all_integration.py
kiddyboots216: Done

tests/test_communication_manager.py
```diff
@@ -255,7 +104,6 @@ def test_communication_manager_can_inform_new_job_to_the_optimizer():
     assert optimizer_job.hyperparams == true_job.hyperparams
     assert optimizer_job.label_column_name == true_job.label_column_name
 
-
 # NOTE: The following are tests that we will implement soon.
```
neeleshdodda44: Update on this?

tests/test_all_integration.py
kiddyboots216: Done

tests/test_all_integration.py
kiddyboots216: Done

tests/test_all_integration.py
kiddyboots216: Ah that actually works very well. Note: in future, fixtures should be session scope by default. This speeds up the tests a LOT.

core/runner.py
kiddyboots216: Okee.

tests/test_communication_manager.py
kiddyboots216: Er....haha!

tests/test_all_integration.py
neeleshdodda44: Can you refactor relevant fixtures in optimizer/communication_manager tests so that they are session scope too? Too lazy to comment on all of those.

Also nice catch on scope, I actually assumed the default behavior was module the entire time.

tests/test_communication_manager.py
neeleshdodda44: Nvm lol. Resolving.

tests/test_all_integration.py
kiddyboots216: Done (tests are now 45 seconds faster!)

core/runner.py

kiddyboots216:
Fixed the bug with Scheduler; it was using collections.deque which is not multiprocessing-safe or thread-safe.

PR #34, opened 2018-11-15
Optimizer transform split

kiddyboots216:
This PR integrates TRANSFORM_SPLIT into the Fed. Avg. Optimizer by updating the Optimizer to return an array of Jobs, updating the Communication Manager to run that array of Jobs, and updating all the tests so that they are ok with this new behavior.
It's worth noting that the transform_function absolutely should not be on master. But that should be addressed in a separate PR if and when we make the decision to remove it; for now, I've added a few lines in _transform_and_split_job in the Runner to handle that.


georgymh:
Minor things:

- Can we return ActionableEventTypes.NOTHING.name, None instead of ActionableEventTypes.NOTHING.name, self.job when appropriate within the Optimizer?

- Your changes in the runner seem correct. Do you want to just remove the transform_function properrt from the DMLJob and the tests to complete Neel's Patch PR here? (Otherwise expect to resolve merge conflicts; we can do this here because it's short.)

- Can you rename ActionableEventTypes.SCHEDULE_JOB to its plural form?

- FYI, within the Optimizer's kickoff method, you could have just used a copy constructor for the 2 DMLJob you created for simplicity.

- The changes you made to the gitignore are OK, but there's a cleaner way to prevent those files from being committed and that's just cleaning them up after the tests. Could you do that instead?

I didn't thoroughly check the test files changed by this PR but besides what I mentioned above, I think the other files look good. Neel, if you're able to, please review the tests for Panda.

kiddyboots216:
- Done
- Done
- Done
- Considered this. Didn't want to add more bloat to DMLJob. If I ever have to do it again I will add the copy constructor to DMLJob class.
- I guess, but rm -ring that stuff is pretty annoying...the point of .gitignore is to prevent people who don't know what they're doing from committing incorrectly. Sure if everyone working on the repo understands to rm -r all those files in all those directories every time.

kiddyboots216:
@georgymh if you can approve the patch or give all necessary by tmrw 4:30 PM that would be ideal, as I'll have another few hours to work tomorrow and if this is merged I can get communication-job ready for review.

georgymh:
With the last point I meant automating the clean up within the tests.

I have some time after class. I will try.

core/communication_manager.py
```diff
@@ -102,15 +102,16 @@ def _create_session(self, payload):
         callback(payload)
         logging.info("Kickoff complete.")
 
-    def _schedule_job(self, dmljob_obj):
+    def _schedule_job(self, dmljob_obj_arr):
```
neeleshdodda44: Nitpick but s/_schedule_job/_schedule_jobs

core/runner.py
```diff
@@ -284,12 +281,11 @@ def _transform_and_split_data(self, job):
...
         os.makedirs(session_filepath)
 
-        # 4. Retrieve transform function and train-test split from job.
-        transform_function = job.transform_function
+        # 4. Retrieve train-test split from job.
         split = job.hyperparams['split']
 
         # 5. Shuffle raw data, then split into train and test set.
-        transformed_data = transform_function(raw_data)
+        transformed_data = raw_data
```
neeleshdodda44: Since there's no more transform function, you should just delete this line and s/transformed_data/raw_data since it would be misleading otherwise.

tests/test_communication_manager.py
```diff
@@ -250,7 +250,7 @@ def test_communication_manager_can_inform_new_job_to_the_optimizer():
     )
     optimizer_job = communication_manager.optimizer.job
     assert optimizer_job.weights == true_job.weights
-    assert optimizer_job.job_type == true_job.job_type
+    # assert optimizer_job.job_type == true_job.job_type
```
neeleshdodda44: Why was this line commented out? Was it intentional?

kiddyboots216: Yes, it's intentional. We shouldn't be checking anything based on the input jobtype as that shouldn't matter.

core/runner.py

core/communication_manager.py
```diff
@@ -102,15 +102,16 @@ def _create_session(self, payload):
...
         Helper function that schedules a DML Job through the Scheduler.
         """
         if not self.scheduler:
             raise Exception("Communication Manager needs to be configured first!")
-        if not isinstance(dmljob_obj, DMLJob):
-            raise Exception("Job is not valid!")
-        self.scheduler.add_job(dmljob_obj)
+        for dmljob_obj in dmljob_obj_arr:
+            if not isinstance(dmljob_obj, DMLJob):
+                raise Exception("Job is not valid!")
```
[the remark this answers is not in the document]
kiddyboots216: 1. Done
2. Done

core/communication_manager.py
kiddyboots216: Done

tests/test_communication_manager.py
neeleshdodda44: Okay then I think you should just go ahead and delete this line.

core/communication_manager.py
```diff
@@ -102,15 +102,16 @@ def _create_session(self, payload):
...
         Helper function that schedules a DML Job through the Scheduler.
         """
         if not self.scheduler:
             raise Exception("Communication Manager needs to be configured first!")
-        if not isinstance(dmljob_obj, DMLJob):
-            raise Exception("Job is not valid!")
-        self.scheduler.add_job(dmljob_obj)
+        for dmljob_obj in dmljob_obj_arr:
+            assert isinstance(dmljob_obj, DMLJob), \
+                "Is not DMLJob, is {}".format(type(dmljob_obj))
```
neeleshdodda44: Okay sorry I wasn't clear enough. This new information is good, but what I meant was that it's difficult to tell which of the jobs in the array is not the correct type. For example, if there are 100 jobs in the array, but one of them is actually isn't a DMLJob (for whatever reason). How would you figure out which of the jobs was the fake DMLJob? 

I get that this is just defensive programming, and that the probability we even run into such a situation is low, but in case we do, I think this would save a lot of time.

Tl;dr I would add an index or some kind of identifier so that if this assertion fails, it isn't difficult to figure out which job in the array the assertion failed on.

core/fed_avg_optimizer.py
```diff
@@ -78,8 +78,32 @@ def kickoff(self):
...
+			framework_type=self.job.framework_type,
+			weights=self.job.weights,
+			hyperparams=self.job.hyperparams,
+			label_column_name=self.job.label_column_name,
+			raw_filepath=self.job.raw_filepath,
+			session_filepath=self.job.session_filepath,
+			datapoint_count=self.job.datapoint_count
+		)
+		job_arr.append(job_one)
+		job_two = DMLJob(
```
neeleshdodda44: Two things.

1. It seems like you're assigning all of the attributes for both the INIT and TRANSFORM_SPLIT job. Why? For INIT, you shouldn't need the session_filepath or datapoint_count (you already know they'll both be `None` since it's coming from `initialization_payload`. Same for TRANSFORM_SPLIT, since the result of that job will assign those attributes. I don't think you assign attributes in jobs that won't be used in the job itself, it's kind of misleading as to what the job needs (the default value for those is `None` anyway). 

2. This might seem trivial but maybe we should implement a feature of DMLJob to create a hard copy of itself (and attributes) and set the other parameters manually. Or a another DMLJob constructor that takes another DMLJob as an argument and sets the remaining arguments as the remaining attributes. For example:
```
job_two = DMLJob(
          job=job_one,
          job_type=JobTypes.JOB_INIT.name
)
 ```
Just seems repetitive to have to initialize a job from scratch if you're creating the job from another job (most of the attributes you're directly copying anyway).

File: core/fed_avg_optimizer.py [hunk 7]
```diff
@@ -105,12 +129,30 @@ def _handle_job_done(self, dmlresult_obj):
...
 		"LEVEL 2" Callback for a initialization job that just completed. Returns
-		a DML Job of type train and modifies the current state of the job.
+		a DML Job of type train unless transforming and splitting is not yet done
+		and modifies the current state of the job.
 		"""
 		new_weights = dmlresult_obj.results.get('weights')
 		self._update_weights(new_weights)
-		self.job.job_type = JobTypes.JOB_TRAIN.name
-		return ActionableEventTypes.SCHEDULE_JOB.name, self.job
+		if self.job.job_type == JobTypes.JOB_TRAIN.name:
```
neeleshdodda44: So if I understand correctly, you're scheduling an INIT and TRANSFORM_SPLIT job at the same time. When the first job finishes, you set `self.job.job_type = JobTypes.JOB_TRAIN.name`. That way when the other job finishes, it knows the other one already finished and you can start training.

If so, this is a pretty good solution. However I think you should just use separate variables that can be booleans (set to False at start). For example, `_done_initializing` can check a variable called `self.has_finished_transform_split`. If the variable is True, then start training, otherwise set `self.has_finished_initializing` to True.  Vice versa for other method. This might seem nitpicky, but I think it makes your algorithm a lot more readable and easier to understand.

tests/test_dmlscheduler.py
```diff
@@ -90,3 +94,12 @@ def test_dmlscheduler_cron():
         initial_weights = output.results['weights']
         assert type(initial_weights) == list
         assert type(initial_weights[0]) == np.ndarray
+
+# def test_dmlscheduler_cannot_schedule_transform_and_split_jobs(mnist_filepath):
```
neeleshdodda44: If this has been resolved, you can go ahead and remove it?

tests/testing_utils.py
```diff
@@ -63,16 +63,16 @@ def make_train_job(model_json, initial_weights, hyperparams, \
...
-    initialize_job = make_initialize_job(make_model_json())
-    if session_filepath:
-        datapoint_count = count_datapoints(session_filepath)
-        initialize_job.datapoint_count = datapoint_count
-        initialize_job.session_filepath = session_filepath
+    initialize_job = make_initialize_job(make_model_json(), raw_filepath)
+    # if session_filepath:
+    #     datapoint_count = count_datapoints(session_filepath)
+    #     initialize_job.datapoint_count = datapoint_count
+    #     initialize_job.session_filepath = session_filepath
```
neeleshdodda44: You should just go ahead and remove this block. Also, please remove the TODO in the docstring/

tests/test_dmlscheduler.py
kiddyboots216: It's not resolved really -if we add the transform function back in, this test will still be useful to show us that multiprocessing can't handle transform_and_split.

tests/test_communication_manager.py
kiddyboots216: Done

core/communication_manager.py
kiddyboots216: Done (see enumerate)

tests/test_dmlscheduler.py
neeleshdodda44: Wait why? If this is the Pickle Error, we can just convert the lambda function to a regular function.

neeleshdodda44: And regardless, I don't think we'll be passing transform functions back and forth anyway per Georgy's reasoning.

kiddyboots216: Ok, removed

tests/testing_utils.py
kiddyboots216: Done

core/fed_avg_optimizer.py
kiddyboots216: Separate variables encounter synchronization issues. Recall that _done_initializing and _done_training are by definition running in different processes. If _done_initializing hits the 'else' and then is preempted by _done_training, which also hits the 'else' because _done_initializing was preempted before it set the variable to True, then we have an issue.
When two processes in Python handle the same variable there's an implicit thread_lock being acquired and released.
See https://en.wikipedia.org/wiki/Lock_(computer_science) for more details on locks.
I can use another variable name, or add to the docstring, to make it easier to understand.

core/fed_avg_optimizer.py
kiddyboots216: Done

core/fed_avg_optimizer.py
neeleshdodda44: I see but could you explain why the same issue doesn't apply with the job type? Specifically, what if both processes hit the else clause before the job type is updated?

kiddyboots216: >When two processes in Python handle the same variable there's an implicit thread_lock being acquired and released.

neeleshdodda44: But I'm asking about the case where the first process will check the job type -> move to else, then the second process will check the job type -> move to else, first process sets job type, second process sets job type.

neeleshdodda44: Doesn't the lock get released when one process is done with whatever clause/assignment uses that variable?

kiddyboots216: That case doesn't occur because both functions acquire any locks for variables that they are setting at the beginning of the function. So the second process is blocked if both of them are setting the same variable, since it needs that lock. However if they aren't using the same variable, Python sees no need to enforce the lock - after all, if you acquired a lock just for reading data, multiprocessing would be pretty bad.

kiddyboots216: I'm not 100% sure on this stuff, this is just what I gathered from reading the multiprocessing documentation and experimentation. We probably want to use an actual Lock() primitive, just not sure where it would get instantiated/used and I don't want to deal with that rn.

neeleshdodda44: Ah I see. So if you set something to both variables, in each function, then the problem wouldn't occur (not that I want you to, just trying to understand).

Okay so one variable works. Can you still not use the job type and still use another variable (maybe a new variable called `first_job_finished`)?

Also just going to approve now since everything else looks good.

PR #35, opened 2018-11-17
Dataset mappings

neeleshdodda44:
Hopefully, this a short PR. 

**Dataset Manager**
- Added a method `create_dataset_mappings` to generate mappings from random UUIDs to dataset folders, and store this in a yaml file called `datasets.yaml`.
- Added a method `bootstrap` to decide whether to load the yaml file (if it exists) or call `create_dataset_mappings`
- Added a method `get_mappings` that returns the loaded mappings, if they exist.
- Added tests for these new methods.

Tl;dr the implementation mirrors that of the Configuration Manager (I ultimately decided not to just put the mappings in configuration.ini since I thought that would break abstractions).

NOTE: I accidentally committed parts of the first two methods above in a previous PR, please make sure to still review all of each of the methods.


core/dataset_manager.py
```diff
@@ -54,7 +55,7 @@ def __init__(self, config_manager):
         raw_filepath = config['GENERAL']['dataset_path']
         assert os.path.isdir(raw_filepath), "The dataset filepath provided is not valid."
         self.raw_filepath = raw_filepath
-        self.mappings = None
+        self._mappings = None
```
kiddyboots216: It's good that this is private. Can you make everything that should be private in datasetManager private?

core/dataset_manager.py
```diff
@@ -107,24 +108,51 @@ def _validate_data(self):
...
+        Returns True if mappings had to be created, and False otherwise 
+        (used for testing).
+        """
+        mapping_filepath = os.path.join(self.raw_filepath, "datasets.yml")
         if os.path.isfile(mapping_filepath):
-            with open(mapping_filepath, "r") as f:
-                self.mappings = yaml.load(f.readlines())
+            if not self._mappings:
+                with open(mapping_filepath, "r") as f:
+                    self._mappings = yaml.load(f.readlines())
```
kiddyboots216: Since this will always call _create_dataset_mappings() which will always set the mappings, what's the point of doing this here? Is it the same or different? They should probably both be in one place. If it's duplicated in both places, I get that it's "defensive programming" but I also think this would make it harder to spot any errors. This method should assume that _create_dataset_mappings does exactly what it's supposed to do, and either call it or not call it but not try to do part of its job.

neeleshdodda44: Can you clarify what you mean? _create_dataset_mappings is only called when the `datasets.yaml` doesn't exist.

kiddyboots216: On startup, datasets.yaml doesn't exist, right.
Then _create_dataset_mappings is called which sets self._mappings
So why is this again setting self._mappings?

core/dataset_manager.py
kiddyboots216: also you forgot to change the tests you scrub

core/dataset_manager.py
neeleshdodda44: Oh, the mappings have to stay consistent. It'll make more sense if I explain with an example.

When data provider's ED Directories are pushed, identifiers are used to represent the datasets. Then the Notebook will use these identifiers to tell the service what datasets the model developer wants to train on. So let's say the service stopped running on the data provider's machine for whatever reason so that the mappings aren't loaded when it starts again. The mappings can't be regenerated (random strings), so we need to use the same mappings.

Tl;dr identifiers are non-deterministic so they need to be saved so the service knows what datasets to train on at a later point.

kiddyboots216: Ok makes sense thanks for example.

core/dataset_manager.py
neeleshdodda44: heckin bamboozled

kiddyboots216: ok now ur actually failing tests

neeleshdodda44: no it was more stuff I forgot to update -_-

to quote a good friend of mine:
> I'm too high from all this berkeley smoke

neeleshdodda44:
@georgymh could you explain why UUID again? Like instead of just using the dataset names or other ways of generating random identifiers?

I have an idea of how to make this smoother, but I'm not sure if it would compromise your original reasons for having an identifier in the first place.

neeleshdodda44:
So the reason why I ask is because it might be better to use some kind of deterministic encoding (like uuid3/uuid5 and some arbitrary namespace in the [docs](https://docs.python.org/3/library/uuid.html)) so that we don't even need to store the mappings in some file. Obviously not necessary (and there's some drawbacks), but I think it's alternative to consider.

Since I've got two approvals, I'm just going to go ahead and merge.

PR #36, opened 2018-11-18
Config Manager clean up + fixing bug

neeleshdodda44:
I made this PR in response to some issues in #31 that @kiddyboots216  was having with adding another secretion to configuration.ini. I found a really stupid bug that only occurs with 3 or more secret sections (I would go into detail, but it's kind of complicated to explain).

I also cleaned up the tests so that everything is easier to understand and so that the tests only pull questions from artifacts.


tests/artifacts/config_manager/questions2.csv
```diff
@@ -0,0 +1,20 @@
+question,section,key,default
```
kiddyboots216: do we need questions2?

kiddyboots216: nvm

neeleshdodda44: Technically no, but I wanted modularity in the tests so that we could test whether secret sections don't work period or whether only multiple secret sections work.

questions has only secret section, questions2 has two

PR #37, opened 2018-11-24
Lotion Update

nzoghb:
Update to the `.travis.yml` to take into account the latest changes to `lotion`. Experienced numerous non-trivial bugs, needing multiple different tries; hence the messy commit history.


nzoghb:
@kiddyboots216 not sure what you mean. Is npm not being cached? Maybe I missed it.

PR #38, opened 2018-12-18
Integrate Dataset Manager

neeleshdodda44:
Couple major changes:
1. Bootstrapper creates the dataset mappings when it initializes the Dataset Manager.
2. Optimizer uses Dataset Manager (passed from Bootstrapper -> Communication Manager -> Optimizer) in order to get dataset mappings whenever it gets initialization payload
3. Optimizer receives a uuid on initialization payload instead of the actual filepath. 
4. It uses the mappings to map this uuid to a specific dataset filepath and it sets the filepath to the job.
5. Error handling for uuid
7. Tests and testing utils have been refactored all over to make these changes (might make the PR look long, but there was honestly a lot of stuff to update).

I might need to update documentation and clean up some code (my fault, I took a break on this bc finals and need to review everything a few more times), but the core logic, tests should stay the same and tests pass so it's pretty much ready for review..


core/fed_avg_optimizer.py
```diff
@@ -56,7 +56,12 @@ def __init__(self, initialization_payload):
 		logging.info("Setting up Optimizer")
 		serialized_job = initialization_payload.get('serialized_job')
 		self.job = deserialize_job(serialized_job)
+		mappings = dataset_manager.get_mappings()
+		assert self.job.uuid, "uuid of job not set!"
+		assert self.job.uuid in mappings, "uuid not found in mappings"
+		self.job.raw_filepath = mappings[self.job.uuid]
```
kiddyboots216: So over on explora, we were talking about how each job will need to have a list of mappings from dataset_uuid to label_column_name, and each service will need to look its name up in this mapping. I think it's best if we add this functionality into this PR directly, if we are going to have it. So instead of self.job.uuid we'd need to have that lookup process.

core/utils/dmljob.py
```diff
@@ -63,6 +64,7 @@ def __init__(
         self.omega = None
         self.sigma_omega = None
         self.new_weights = None
+        self.uuid = uuid
```
kiddyboots216: Same as above -doesn't seem like a job is going to have just 1 uuid, it will have a list of mappings.

core/fed_avg_optimizer.py
neeleshdodda44: Sorry, I'm a bit confused. I think we have several misunderstandings.

 1. Why are there mappings from dataset_uuid to label_column_name? There are only mappings from dataset_uuid to dataset_filepath. The label_column_name should already be set when the job is created in explora.

2. Dataset mappings are stored in each service, not each job. That's why jobs only have a uuid, so that each service will perform the lookup to determine which dataset the job is referring to.

 > So instead of self.job.uuid we'd need to have that lookup process.
3. Not sure what you mean. The serialized job passed in initialization payload should have the uuid set at this point, and the lookup is performed with `mappings`

core/utils/dmljob.py
neeleshdodda44: Same as above, jobs should only have 1 uuid (each individual dataset has its own job)

core/fed_avg_optimizer.py
kiddyboots216: This is going off your comment on Explora PR: "Or actually I think a better alternative to 1. is to actually make the nested dictionary a list of tuples, where each tuple is (dataset_uuid, label_column_name). That way, a UNIX Service on someone's machine simply needs to lookup their own name to immediately get the list of datasets to train on."
I'm assuming the lookup process will occur here in the optimizer, as it's the first place that an intelligent component of the service is exposed to the job. So the uuid won't be set at this point, the optimizer will need to do a lookup.

neeleshdodda44: 1. I thought Optimizer only handles one job (one dataset) at a time. This lookup should be in a different component, which receives a list of tuples and creates serialized jobs -> separate messages to Communication Manager. This might require more changes, since we only support one session at a time (only one optimizer).

2. The above changes should really be in a separate PR, since this one is just for integrating the Dataset Manager.

neeleshdodda44: What I meant was that the initial "job" sent from explora that's received by this data provider will have a dictionary of data providers mapping to lists of tuples. By looking up their own name, the data provider will have the tuples of datasets. But work still needs to be done to turn each tuple along with other information sent in the "job" into a formal INIT DMLJob to be handled by the Communication Manager.

kiddyboots216: Ok, there's no point in arguing this. It will come up in the explora-integration PR I guess.

core/communication_manager.py
```diff
@@ -90,9 +94,14 @@ def _create_session(self, payload):
         # NOTE: We removed the 'optimizer_type' argument since for the MVP we're
         # only considering the 'FederatedAveragingOptimizer' for now.
         # TODO: We need to incorporate session id's when we're ready.
+        if not self.dataset_manager:
+            raise Exception("Communication Manager needs to be configured first!")
```
kiddyboots216: I think this is the same Exception as the Exception raised by Comm Mgr didn't have a Scheduler? Can you differentiate them.

tests/test_fed_avg_optimizer.py
```diff
@@ -26,16 +27,26 @@ def config_manager():
     return config_manager
 
 @pytest.fixture(scope='session')
-def initialization_payload(small_filepath):
+def mnist_uuid():
+    return 'd16c6e86-d103-4e71-8741-ee1f888d206c'
```
kiddyboots216: So in future, this is how we are going to be making tests -with the uuid stuff?

PR #39, opened 2018-12-25
Integrated DB Client + Cleaned up ED Directory Push

neeleshdodda44:
Several changes with pushing ED Directories:
- Config Manager asks the user to label the data, which is put in the config 
- Dataset Manager pulls label from the config
- To keep design simple, tuples of (dataset_sample, dataset_metadata) are pushed for each folder in the ED Directory
- DB Client adds label for data provider on ED Directory push
- Added test for DB Client back in (test seems to always pass for now?)
- Added test for checking format of ED Directory

TODOs:
- We need to eventually have a classifier that labels data automatically
- explora currently assumes that each folder has just the sample (and not the tuple of sample,metadata) so this needs to be updated
- explora currently extracts metadata from the samples of data, when it should now use this metadata.


core/blockchain/blockchain_utils.py
```diff
@@ -161,4 +161,4 @@ def __init__(self, key: str, value: str) -> None:
         self.value = value
 
     def get_tx(self) -> dict:
-       return {TxEnum.KEY.name: self.key, TxEnum.CONTENT.name: self.value}
+       return {TxEnum.KEY.name: self.key, TxEnum.CONTENT.name: self.value}
```
kiddyboots216: what?

core/dataset_manager.py
```diff
@@ -50,20 +50,28 @@ def __init__(self, config_manager):
 
         TODO: In the event of a invalid dataset path, log a useful message 
         for the user and don't terminate the service
```
kiddyboots216: Is this TODO resolved yet?

core/dataset_manager.py
```diff
@@ -50,20 +50,28 @@ def __init__(self, config_manager):
...
         self._validate_data()
         self._port = config.getint("BLOCKCHAIN", "http_port")
+        self._ipfs_client = None
+        self._db_client = None
+        self.label = config['DATASET_MANAGER']['category']
 
-    def configure(self, ipfs_client):
+    def configure(self, ipfs_client, db_client):
         """
         Sets up IPFS client for _communicate.
```
kiddyboots216: Update docstring

core/dataset_manager.py
```diff
@@ -152,69 +160,68 @@ def _create_dataset_mappings(self, mapping_filepath):
...
+            if not os.path.isdir(folder_path): continue
+            encoding = reverse_mappings[folder]
+            files = list(os.listdir(folder_path))
+            assert len(files) == 1, \
+                "We only support one file per dataset folder!"
+            file_name = files[0]
+            if not file_name.endswith(".csv"): continue
+            filepath = os.path.join(folder_path, file_name)
+            dataset = pd.read_csv(filepath)
+            sample = dataset.sample(frac=0.1)
```
kiddyboots216: Filesize of sample data shouldn't scale sublinearly to size of real data, it should scale log or preferrably constant. You don't need to fix this in this PR.

core/dataset_manager.py
```diff
@@ -152,69 +160,68 @@ def _create_dataset_mappings(self, mapping_filepath):
...
-            file_path = os.path.join(folder_path, file)
-            dataset = pd.read_csv(file_path)
-            md = pd.DataFrame(dataset.describe())
-            sample = dataset.sample(frac=0.1)
-            folder_dict['ds'] = sample.to_json()
-            folder_dict['md'] = md.to_json()
-            value[folder] = folder_dict
-        receipt = setter(client=self._client, key=name, value=value, port=self._port)
+        assert len(key) <= 30, \
+            "Keys for datasets can only be at most 30 characters long."
```
kiddyboots216: Why is this? I don't think this is a restriction that we have for the blockchain. Is this just an arbitrary length? I'm also not sure what the "dataset key" is.

core/dataset_manager.py
```diff
@@ -152,69 +160,68 @@ def _create_dataset_mappings(self, mapping_filepath):
...
+        assert self._db_client, \
+            "DB client has not been set. Dataset Manager needs to be configured!"
+        assert self._ipfs_client, \
+            "IPFS client has not been set. Dataset Manager needs to be configured!"
+        
+        ed_directory = self._generate_ed_directory()
+
+        self._db_client.add_labels([key], [self.label])
```
kiddyboots216: Why is add_labels called here? Not too sure why the db_client should be doing something in the method that pushes to the blockchain.

core/questions.csv
```diff
@@ -7,6 +7,7 @@ What is the max number of times a job should run?,SCHEDULER,max_tries,3
 What is the port for IPFS?,BLOCKCHAIN,ipfs_port,5001
 What is the port for HTTP?,BLOCKCHAIN,http_port,3000
 What is the host for HTTP?,BLOCKCHAIN,host,127.0.0.1
+What category would you label this data,DATASET_MANAGER,category,default_category
```
kiddyboots216: nitpick, add '?'

tests/test_dbclient.py
```diff
@@ -59,30 +59,30 @@ def reset(db_client, data_provider_list):
     )
 
 # Commenting out tests now due to nondeterministic bugs.
```
kiddyboots216: Ok so -are these tests now going to always pass? Because the TODOs about nondeterministic bugs are still there :o

tests/test_dataset_manager.py
```diff
@@ -132,6 +132,25 @@ def test_no_yaml_creation_repeat(good_dataset_manager):
...
+    ed_directory = good_dataset_manager._generate_ed_directory()
+    filepath = os.path.join(good_dataset_manager._raw_filepath, 'datasets.yaml')
+    mappings = good_dataset_manager._mappings
+
+    assert type(ed_directory == dict), \
+        "ED Directory is not a dictionary!"
+
+    for encoding, data_tuple in ed_directory.items():
+        dataset = data_tuple[0]
+        assert encoding in mappings.keys(), \
```
kiddyboots216: So looks like we're just checking that every item in the encoding is in mappings.keys() shouldn't we also be checking that every item from mappings.keys() that we expect to be in the encoding is there?

kiddyboots216: Nvm

core/blockchain/blockchain_utils.py
neeleshdodda44: idk either

core/dataset_manager.py
neeleshdodda44: No

core/dataset_manager.py
neeleshdodda44: Noted, thanks

core/dataset_manager.py
neeleshdodda44: This was a restriction you gave me way back in the first iteration of Dataset Manager. If it's not relevant anymore, then I can remove it.

Also yeah dataset key is a typo, I meant to say data provider key

kiddyboots216: Technically not relevant since Ethereum -> Tendermint but we should leave it in so that we don't break anything.

core/dataset_manager.py
neeleshdodda44: Category labels need to be pushed at the same time as ED Directories, since the labels are used to retrieve the directories later.

But you're right, I'll change the method name to be more clear about this.

tests/test_dbclient.py
neeleshdodda44: Forgot to delete comment, deleted.

I ran the tests like 10 times and they passed each time so I'm just hoping I didn't get baited and that the bug is gone for good.

core/questions.csv
neeleshdodda44: Nice catch, thanks.

core/dataset_manager.py
neeleshdodda44: Done.

neeleshdodda44:
Fair point, I replaced labels -> classifications in documentation and code to make this clear. I'll do the same for DB Client in explora too.

Merging now.

PR #41, opened 2019-01-03
Uniform Initialization of Models from Seed

kiddyboots216:
By compiling layers of the Keras model with the seed argument set for the glorot_uniform initializer, we can ensure that models are initialized with the same weights. If we choose to have a bias initializer, then we would similarly set the seed for the bias initializer, but the bias is initialized to all zeros by default.

Although this capability is inherent in the Unix Service, and this PR just adds a test to ensure it stays that way, we need to make sure in Explora that we either urge developers to add the seed when they create their Model, or do it for them.

### DIFF
--- a/tests/test_dmlrunner.py
+++ b/tests/test_dmlrunner.py
@@ -90,6 +90,16 @@ def train_dmlresult_obj(config_manager, split_dmlresult_obj, init_dmlresult_obj,
                 )
     result = runner.run_job(train_job)
     return result
+def test_dmlrunner_uniform_initialization(config_manager, ipfs_client):
+    runner = DMLRunner(config_manager)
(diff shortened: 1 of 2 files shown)


tests/test_dmlrunner.py
```diff
@@ -90,6 +90,16 @@ def train_dmlresult_obj(config_manager, split_dmlresult_obj, init_dmlresult_obj,
                 )
     result = runner.run_job(train_job)
     return result
+def test_dmlrunner_uniform_initialization(config_manager, ipfs_client):
```
neeleshdodda44: Can you add an extra line above this line?

kiddyboots216: yeet

PR #42, opened 2019-01-07
Clients do not hear messages they have posted to the blockchain

kiddyboots216:
Replace the array-based state in BlockchainGateway with a multiprocessing-safe List() object so that DMLRunners can append transactions they submit to the blockchain to the current state of the Gateway. Therefore, the Gateway will not hear these transactions. This implementation was chosen over adding a signature field to the transaction in dagora_chain.

All other updates are consequences of this change, such as the changes to test_all_integration.py, blockchain_utils.py, and blockchain_gateway.py.

Also updates the Bootstrapper as that hadn't been updated in a while.

As the technical details of this PR are interesting, I've included test_state_setting.py to illustrate the minimal skeleton of how it works. This file will be deleted once the PR is approved, before merging.


tests/test_all_integration.py
```diff
@@ -339,26 +351,26 @@ def test_communication_manager_integration(
...
+    # )
+    # assert communication_manager.optimizer.job.job_type == JobTypes.JOB_AVG.name, \
+    #     "Should be ready to average!"
+    # timeout = time.time() + 3
+    # while time.time() < timeout and len(scheduler.processed) == 5:
+    #     # second averaging job
+    #     scheduler.runners_run_next_jobs()
+    #     time.sleep(0.1)
+    # assert len(scheduler.processed) == 6, "Averaging failed/not completed in time!"
+    # # we've heard both sets of new weights so our listen_iters are 2
```
neeleshdodda44: Did you intend to leave this commented out?

kiddyboots216: Yep. This indicates the portion of the communication manager test that has been removed because of the state-setting feature, which I think shows the consequence on the actual algorithm the best.

tests/test_dmlscheduler.py
```diff
@@ -35,21 +38,52 @@ def mnist_filepath():
...
+    return blockchain_gateway
+
+@pytest.fixture
+def dataset_manager(config_manager):
+    dataset_manager = DatasetManager(config_manager)
+    dataset_manager.bootstrap()
+    return dataset_manager
 # NOTE: This is scope=function and not scope=session because of multiprocessing.
 # If it's scope=session, tests will get interrupted by each other and then the assertions
 # will fail because more jobs will be processed than they expected.
```
neeleshdodda44: This note only applies for scheduler, right? Could you make the above fixtures scope=session too?

tests/test_all_integration.py
neeleshdodda44: I think you should leave a note of that right before the commented block or honestly just remove it (I would still see the removal either way, but this way the intended impact of the design decision is clearer here)

core/runner.py
```diff
@@ -78,7 +81,10 @@ def run_job(self, job):
             self.JOB_CALLBACKS,
         )
         try:
-            results = callback(job)
+            if job.job_type in self.JOBS_NEEDING_STATE:
+                results = callback(job, state)
+            else:
+                results = callback(job)
```
neeleshdodda44: Should

core/runner.py
```diff
@@ -57,14 +57,17 @@ def __init__(self, config_manager):
             JobTypes.JOB_AVG.name: self._average,
             JobTypes.JOB_COMM.name: self._communicate,
         }
+        self.JOBS_NEEDING_STATE = {
+            JobTypes.JOB_COMM.name: self._communicate,
+        }
```
neeleshdodda44: In line 79, the callback method always comes from the callbacks dictionary. You only use this dictionary to check if a key is in it. Thus, I think you should just make this an array so that's clearer.

tests/test_all_integration.py
```diff
@@ -43,11 +43,11 @@ def ipfs_client(config_manager):
 
 @pytest.fixture(scope='session')
 def mnist_uuid():
-    return 'd16c6e86-d103-4e71-8741-ee1f888d206c'
+    return '0fcf9cbb-39df-4ad6-9042-a64c87fecfb3'
 
 @pytest.fixture(scope='session')
 def mnist_uuid_two():
-    return '0fcf9cbb-39df-4ad6-9042-a64c87fecfb3'
```
neeleshdodda44: You flipped the uuids here and dataset paths in the integration configs. Why?

tests/test_all_integration.py
```diff
@@ -114,37 +129,36 @@ def test_federated_learning_two_clients_automated(
...
-    tx_receipt_two = setter(
+    tx_receipt = setter(
         client=blockchain_gateway._client,
         key=new_session_key_two, 
         port=blockchain_gateway._port,
         value=new_session_event, 
         flag=True
     )
     assert tx_receipt
-    assert tx_receipt_two
```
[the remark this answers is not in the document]
kiddyboots216: Resolved

tests/test_all_integration.py
```diff
@@ -191,50 +205,49 @@ def test_federated_learning_two_clients_manual(
...
         "Jobs {} failed/not completed in time!".format([
             result.job.job_type for result in scheduler.processed])
-    assert len(scheduler_2.processed) == 8, "Jobs failed/not completed in time!"
-    # (4) Gateway_1 listens for the new weights 
+    assert len(scheduler_2.processed) == 7, "Jobs failed/not completed in time!"
+    # (4) Gateway_1 listens for the new weights and hears only 2's weights
     blockchain_gateway._listen(blockchain_gateway._handle_new_session_info,
         blockchain_gateway._filter_new_session_info)
-    # (6) Gateway_2 listens for the new weights
+    # (6) Gateway_2 listens for the new weights and hears only 1's weights
```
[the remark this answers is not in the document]
kiddyboots216: Done

core/runner.py
kiddyboots216: Thanks

tests/test_dmlscheduler.py
kiddyboots216: Thanks

tests/test_all_integration.py
kiddyboots216: Was planning to remove this upon approval of the PR like the other test I mentioned in the opening note of the PR anyways, yep

PR #45, opened 2019-01-27
Refactored DMLJob

kiddyboots216:
DMLJob now has an instance attribute specific_job which is a container class that contains the specific information needed for whichever DMLJob the Optimizer needs the Runner to run. The Optimizer no longer has a self.job instance attributes, but rather a dictionary of job_data.


core/utils/dmljob.py
```diff
@@ -39,102 +32,231 @@ def __init__(
...
+        Args:
+            job_type (str): the job type.
+            specific_job (obj): the specific job container class
+        """
+        self.job_type = job_type
+        self.specific_job = specific_job
+    
+    def copy_constructor(self):
+        newjob = deepcopy(self)
+        return newjob
```
neeleshdodda44: I feel like this is a REALLY good use case for inheritance.

1. specific_job isn't really necessary anymore (just use `type(job)`)?
2. I think the only reason why having a DMLJob still here is so that Runner can verify that all jobs it s takes are DMLJobs. That might need to be refactored to isinstance, but I would argue it's better to not create any DMLJobs directly like you have above.
3. job_type might not actually be necessary anymore either, but that's up to you. Correct me if I'm wrong, but I think comparing class names basically replaces that functionality (i.e. checking if a job is X Job in tests, etc.)

I'd like to hear your thoughts.

tests/test_all_integration.py
```diff
@@ -359,7 +357,7 @@ def test_communication_manager_integration(new_session_event, new_session_key, c
     assert communication_manager.optimizer.curr_averages_this_round == 0, \
         "Either did not hear anything, or heard too much!"
     # now we should be ready to train
-    assert communication_manager.optimizer.job.job_type == JobTypes.JOB_TRAIN.name, \
-        "Should be ready to train!"
+    # assert communication_manager.optimizer.job.job_type == JobTypes.JOB_TRAIN.name, \
+    #     "Should be ready to train!"
```
neeleshdodda44: Why are all these type checks commented?

kiddyboots216: These type checks worked because the optimizer had a job instance attribute which essentially "forecasted" what the optimizer was going to do. However, with this refactor the optimizer is more or less stateless so it doesn't maintain "intent" (optimizer.job.job_type)

core/utils/dmljob.py
kiddyboots216: So you basically want all the specific DML Jobs to inherit from DMLJob, like DMLAverageJob(DMLJob)? Sure.

tests/test_all_integration.py
neeleshdodda44: Sorry sorry I meant commented out vs deleted. Like why not delete them?

core/utils/dmljob.py
neeleshdodda44: Yes.

tests/test_all_integration.py
kiddyboots216: It's still possible to check these, technically. It's just very slow -so these are left in so that someone can know what they are supposed to be looking for if they want to check.

core/utils/dmljob.py
kiddyboots216: Done

tests/test_all_integration.py
neeleshdodda44: Okay sounds good

core/utils/dmljob.py
neeleshdodda44: Looks good! Feel free to merge

PR #48, opened 2019-02-01
Status server post

kiddyboots216:
Adds a new Optimizer, CloudConnectedOptimizer, which schedules a ServerJob after receiving Training results so that these statistics can be posted to our cloud server.


core/cloud_connected_optimizer.py
```diff
@@ -0,0 +1,50 @@
...
+from core.utils.enums 					import JobTypes, callback_handler_no_default
+from core.utils.dmljob 					import (DMLJob, DMLAverageJob, DMLCommunicateJob, DMLServerJob, \
+												DMLInitializeJob, DMLSplitJob, DMLTrainJob, DMLValidateJob)
+from core.utils.keras 					import serialize_weights, deserialize_weights
+from core.utils.dmlresult 				import DMLResult
+from core.blockchain.blockchain_utils	import TxEnum, content_to_ipfs
+from core.fed_avg_optimizer import FederatedAveragingOptimizer
+
+
+class CloudConnectedOptimizer(FederatedAveragingOptimizer):
```
neeleshdodda44, line 14: 1. Class docstring?

2. Can you summarize all of the differences that this class brings? This can be put in the docstring if you want.

core/cloud_connected_optimizer.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, initialization_payload, dataset_manager):
+        super().__init__(initialization_payload, dataset_manager)
+        job_info = initialization_payload.get(TxEnum.CONTENT.name)
+        serialized_job = job_info.get('serialized_job')
+        self.job_data["uuid"] = serialized_job["uuid"]
```
neeleshdodda44: 1. Aren't lines 17-18 unnecessary because of line 16? `self.job_data` should already be set...?
2. You should really s/uuid/job_uuid everywhere just to emphasize that it's job_uuid (because uuid by itself used to mean dataset_uuid)

core/runner.py
```diff
@@ -282,6 +285,29 @@ def _communicate(self, job, state):
         )
         return results
 
+    def _post_statistics(self, job):
+        """
+        Posts statistics about a training job to a cloud server
+        """
+        assert job.statistics, "No statistics to be served"
+        server_url = "http://status-server.jn6tkty4uh.us-west-1.elasticbeanstalk.com/jobs/{0}/{1}".format(job.uuid, job.dataset_uuid)
```
neeleshdodda44: I think it would be better to move to this to ServerJob, the Runner shouldn't keep track of information regarding the server. 

Within ServerJob, you could honestly just have a global variable (but it would be best to pull from config)

tests/test_dmlrunner.py
```diff
@@ -113,9 +113,15 @@ def runner(config_manager, ipfs_client):
     runner.configure(ipfs_client=ipfs_client)
     return runner
 
-def test_dmlrunner_uniform_initialization(config_manager, ipfs_client):
-    runner = DMLRunner(config_manager)
-    runner.configure(ipfs_client)
+def test_dmlrunner_status_server_post(runner, train_dmlresult_obj):
+    server_job = DMLServerJob(uuid = "pandata", 
+                            statistics=train_dmlresult_obj.results.get('train_stats'),
+                            dataset_uuid = "sucks",
```
neeleshdodda44, line 119: LMAO

core/cloud_connected_optimizer.py
kiddyboots216: 17-18 are just getting serialized_job from the initialization_payload so that line 19 can go off without a hitch. This optimizer requires a job_uuid. Will do for point 2).

core/runner.py
kiddyboots216: I was considering that but this info will change based on the config and there's no way to change the jobs based on the config-? I put it in runner bc runner can actually get info from config

core/cloud_connected_optimizer.py
kiddyboots216: Truuuu. Will do.

core/cloud_connected_optimizer.py
neeleshdodda44: Ah I see, serialized_job isn't saved as a class attribute in the superclass

core/runner.py
neeleshdodda44: Sure, so can you move the url to the config and have Runner pull from the config? This would probably be a secret section.

core/cloud_connected_optimizer.py
kiddyboots216: 2) Done

core/cloud_connected_optimizer.py
kiddyboots216: Done

core/cloud_connected_optimizer.py
```diff
@@ -0,0 +1,58 @@
...
+from core.fed_avg_optimizer import FederatedAveragingOptimizer
+
+
+class CloudConnectedOptimizer(FederatedAveragingOptimizer):
+    """
+    This is a clone of FedAvgOptimizer, but for two differences.
+    1) This Optimizer requires the job_uuid field to be set in
+    its initialization payload
+    2) When this is finished training, it uploads statistics to
+    a centralized cloud repo
```
neeleshdodda44, line 20: I like the idea of using inheritance here, but I have some thoughts.
1. I disagree that only this class should have the job_uuid field, because every optimizer should need the job_uuid in order to relay information back to Explora, right?
2. I like that posting statistics is sort of separate from the core functionality of the optimizer. But I don't understand why this functionality is in a subclass vs a new component. Is there any reason why we would still use the old optimizer over this one?

neeleshdodda44 (approved):
Good work! Feel free to merge.

